PR #8, opened 2024-08-21
Map netflix to wikidata

cocomittens:
My attempt at this task that somewhat works. 

Works for both movies and TV shows. The abomination of explicitly defining various different types of subcategories of these is there as it increases the found rate. Even though from what I understand it should be able to determine subcategories of films (which I would assume would also work for tv shows) without that. But it's there for now because it works.

Outputs `netflix_id`, `wikidata_id`, `year` and `genre` to csv file.
Currently set to the first 500 for testing purposes, also because it crashes trying to do the whole thing on my pc. Which idk if it is specific to my device or some specific movie/ show.

Looking at the data I think there are a few that might not be correct (the ones with genre `None`), and a few edge cases like tv show names that are also movies. Found rate is about ~51% for first 100 and 500. A lot of them actually don't exist from what I've seen (eg `Paula Abdul's Get Up & Dance`), but there are also some that do exist that aren't found (eg `Chain of Command`). Also, if we are doing TV shows, I think anything with `Season` included definitely won't be found on wikidata.

I plan to make it work for all the movies, add another function to specify which subcategories to include in the data, and maybe functionality to save it to the db (or that might be out of scope for this task).


src/data_processing/netflix_to_wikidata.py
```diff
@@ -0,0 +1,110 @@
...
+
+load_dotenv()
+
+base_dir = os.getenv('BASE_DIR')
+user_agent = os.getenv('USER_AGENT')
+
+netflix_data_path = os.path.join(base_dir, 'netflix_movies.txt')
+output_csv_path = os.path.join(base_dir, 'netflix_to_wikidata.csv')
+
+netflix_data = pd.read_csv(netflix_data_path, delimiter=',', names=['NetflixId', 'Year', 'Title'],  encoding='ISO-8859-1', on_bad_lines='skip')
```
audiodude, line 14: Let's use the built in python `csv` library:

https://docs.python.org/3/library/csv.html

src/data_processing/netflix_to_wikidata.py
```diff
@@ -0,0 +1,110 @@
...
+
+def construct_query(title, year):
```
audiodude, line 16: It's great that this is in it's own function!

src/data_processing/netflix_to_wikidata.py
```diff
@@ -0,0 +1,110 @@
...
+    query = '''
+    SELECT * WHERE {
+        SERVICE wikibase:mwapi {
+            bd:serviceParam wikibase:api "EntitySearch" ;
+                            wikibase:endpoint "www.wikidata.org" ;
+                            mwapi:search "%s" ;
```
audiodude, line 22: for a query this big, I would name the params:

```
%(search_term)s
```

just so it's easier to understand in the future.

src/data_processing/netflix_to_wikidata.py
```diff
@@ -0,0 +1,110 @@
...
+            FILTER (lang(?genreLabel) = "en") .
+        }
+
+        ?item rdfs:label ?itemLabel .
+        FILTER (lang(?itemLabel) = "en") .
+
+        SERVICE wikibase:label { bd:serviceParam wikibase:language "en" . }
+    }
+    LIMIT 1
+    ''' % (title, year - 2, year + 2)
```
audiodude, line 62: when you change it to `%(search_term)`, you will have to fill it with a dictionary instead of a tuple.

src/data_processing/netflix_to_wikidata.py
```diff
@@ -0,0 +1,110 @@
...
+    wikidata_id = data['results']['bindings'][0]['item']['value'].split('/')[-1]
+
+    genres = [d['genreLabel']['value'] for d in data['results']['bindings'] if 'genreLabel' in d]
+
+    return wikidata_id, genres
+
+data = []
+missing_count = 0
+
+for index, row in netflix_data.iloc[:500].iterrows():
```
audiodude, line 89: As said, just use `for row in csv_reader:`

src/data_processing/netflix_to_wikidata.py
```diff
@@ -0,0 +1,110 @@
...
+    netflix_id = row['NetflixId']
+    wikidata_id, genres = get_wikidata_data(title, year)
+    if wikidata_id:
+        print(f'found: {title} ({year})')
+        data.append({
+            'netflix_id': netflix_id,
+            'wikidata_id': wikidata_id,
+            'title': title,
+            'year': year,
+            'genres': ', '.join(genres) if genres else 'None'
```
audiodude, line 101: If we're putting this in mongo, it can just be a list.

src/data_processing/netflix_to_wikidata.py
```diff
@@ -0,0 +1,110 @@
...
+        })
+    else:
+        print(f'missing: {title} ({year})')
+        missing_count += 1
+
+df = pd.DataFrame(data)
```
audiodude, line 107: Again, let's just use python csv.

src/data_processing/netflix_to_wikidata.py
```diff
@@ -11,7 +11,18 @@
 netflix_data_path = os.path.join(base_dir, 'netflix_movies.txt')
 output_csv_path = os.path.join(base_dir, 'netflix_to_wikidata.csv')
 
-netflix_data = pd.read_csv(netflix_data_path, delimiter=',', names=['NetflixId', 'Year', 'Title'],  encoding='ISO-8859-1', on_bad_lines='skip')
+netflix_data = []
+
+with open(netflix_data_path, mode='r', encoding='ISO-8859-1') as file:
+    reader = csv.reader(file, delimiter=',')
```
audiodude, line 17: You don't need `delimeter=','`, it's the default

src/data_processing/netflix_to_wikidata.py
```diff
@@ -11,7 +11,18 @@
...
+    for row in reader:
+        if len(row) == 3:
```
audiodude, line 19: As mentioned in Sofia's review, maybe we should just read the file line by line and do `split(',', 2)` in order to get movie titles with commas, as opposed to this strategy which just ignores them.

audiodude (changes requested):
Found a couple more things, nothing major.

src/data_processing/netflix_to_wikidata.py
```diff
@@ -39,7 +39,7 @@ def construct_query(title, year):
         ?item wdt:P31/wdt:P279* wd:Q11424 .
 
         ?item wdt:P577 ?releaseDate .
-        FILTER (YEAR(?releaseDate) >= %d && YEAR(?releaseDate) <= %d) .
+        FILTER (YEAR(?releaseDate) >= %(start_year)d && YEAR(?releaseDate) <= %(end_year)d) .
```
audiodude, line 42: I just tested this in at query.wikimedia.org: https://w.wiki/B4XX

The only problem is that if you use `&& YEAR(?releaseDate <= 2003`, it captures any movie that was released in 2003 (https://w.wiki/B4XX). It looks like you can just use `FILTER (YEAR(?releaseDate) = 2002)` and skip the ranges.

src/data_processing/netflix_to_wikidata.py
```diff
@@ -53,7 +53,7 @@ def construct_query(title, year):
         SERVICE wikibase:label { bd:serviceParam wikibase:language "en" . }
     }
     LIMIT 1
```
audiodude, line 55: If you want to get all of the genres, you need to remove this.

src/data_processing/netflix_to_wikidata.py
```diff
@@ -30,7 +30,7 @@ def construct_query(title, year):
         SERVICE wikibase:mwapi {
             bd:serviceParam wikibase:api "EntitySearch" ;
                             wikibase:endpoint "www.wikidata.org" ;
-                            mwapi:search "%s" ;
+                            mwapi:search "%(search_term)s" ;
                             wikibase:limit 1 ;
```
audiodude, line 34: Remove the limit from the search query so that we get multiple results that we can sift through. Otherwise, you don't get `Lilo and Stitch` the 2002 movie.

src/data_processing/netflix_to_wikidata.py
cocomittens: What about the differences in release date between the dvd and theater release tho? Are we cool with missing those if the dvd version was released in a different year?

src/data_processing/netflix_to_wikidata.py
```diff
@@ -72,7 +72,7 @@ def get_wikidata_data(title, year):
 
     wikidata_id = data['results']['bindings'][0]['item']['value'].split('/')[-1]
 
-    genres = [d['genreLabel']['value'] for d in data['results']['bindings'] if 'genreLabel' in d]
+    genres = {d['genreLabel']['value'] for d in data['results']['bindings'] if 'genreLabel' in d}
```
audiodude, line 75: Did you just use a "set" comprehension? I didn't realize such a thing existed!

cocomittens: Yea I was trying to just eliminate duplicates with the least amount of effort, so just tried to make it a set and found that and it works. Though on second thought I'm not sure if there should be duplicates (if its getting data from other movies or something), so not 100% sure if it's a problem that necessarily should be solved.

audiodude: Makes sense. I think you should "manually" run the query on query.wikidata.org for any movies that are returning duplicate genres. Like you said, it could be that it's getting multiple movies. Or something else could be happening with the query that we need to debug. If you can gather the queries (or movie names/release dates) of such instances, we can debug this Thursday.

cocomittens: I tested it for [Dragonheart](https://query.wikidata.org/#%20%20%20%20SELECT%20%2a%20WHERE%20%7B%0A%20%20%20%20%20%20%20%20SERVICE%20wikibase%3Amwapi%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20bd%3AserviceParam%20wikibase%3Aapi%20%22EntitySearch%22%20%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20wikibase%3Aendpoint%20%22www.wikidata.org%22%20%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20mwapi%3Asearch%20%22Dragonheart%22%20%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20wikibase%3Alimit%201%20%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20mwapi%3Alanguage%20%22en%22%20.%0A%20%20%20%20%20%20%20%20%20%20%20%20%3Fitem%20wikibase%3AapiOutputItem%20mwapi%3Aitem%20.%0A%20%20%20%20%20%20%20%20%7D%0A%0A%20%20%20%20%20%20%20%20%3Fitem%20wdt%3AP31%2Fwdt%3AP279%2a%20wd%3AQ11424%20.%0A%0A%20%20%20%20%20%20%20%20%3Fitem%20wdt%3AP577%20%3FreleaseDate%20.%0A%20%20%20%20%20%20%20%20FILTER%20%28YEAR%28%3FreleaseDate%29%20%3E%3D%201995%20%26%26%20YEAR%28%3FreleaseDate%29%20%3C%3D%201997%29%20.%0A%20%20%20%20%20%20%20%20%0A%20%20%20%20%20%20%20%20OPTIONAL%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%3Fitem%20wdt%3AP136%20%3Fgenre%20.%0A%20%20%20%20%20%20%20%20%20%20%20%20%3Fgenre%20rdfs%3Alabel%20%3FgenreLabel%20.%0A%20%20%20%20%20%20%20%20%20%20%20%20FILTER%20%28lang%28%3FgenreLabel%29%20%3D%20%22en%22%29%20.%0A%20%20%20%20%20%20%20%20%7D%0A%0A%20%20%20%20%20%20%20%20%3Fitem%20rdfs%3Alabel%20%3FitemLabel%20.%0A%20%20%20%20%20%20%20%20FILTER%20%28lang%28%3FitemLabel%29%20%3D%20%22en%22%29%20.%0A%0A%20%20%20%20%20%20%20%20SERVICE%20wikibase%3Alabel%20%7B%20bd%3AserviceParam%20wikibase%3Alanguage%20%22en%22%20.%20%7D%0A%20%20%20%20%7D%0A%20%20%20%20LIMIT%2010%0A) and it appears that there are multiple release dates in different countries which is causing it to be returned 5 times. So there is at least one scenario where it is returning a valid movie but still includes duplicates. Maybe could just change it to only search for US release date or something? I'll look at the other ones to see if there is anything else causing it.

cocomittens: Other duplicates in the first 100 include:
Presumably caused by multiple release dates -
[Mostly Martha](https://www.wikidata.org/wiki/Q684936)
[Silkwood](https://www.wikidata.org/wiki/Q1412037)
[The Killing](https://www.wikidata.org/wiki/Q592072)
[Jingle all the Way](https://www.wikidata.org/wiki/Q1064530)
[Congo](https://www.wikidata.org/wiki/Q598860)
[At Home Among Strangers](https://www.wikidata.org/wiki/Q1963834)
[Richard III](https://www.wikidata.org/wiki/Q2501468)
[Jade](https://www.wikidata.org/wiki/Q889386)
[The Spitfire Grill](https://www.wikidata.org/wiki/Q1212954)
Valid movie, unsure what's causing it - 
The classic [Powerpuff Girls Movie](https://www.wikidata.org/wiki/Q1199301)
Film and television film within same year -
The Love Letter [Film](https://www.wikidata.org/wiki/Q949305), [Television Film](https://www.wikidata.org/wiki/Q3363005)

cocomittens: So yeah basically it seems like the most is due to the release dates issue. Which could probably be solved more efficiently, but technically valid to just remove duplicates.
2 are because of the date range issue. I'll try running it with just the single date and see how much that changes the results.
2 are unknown reasons.

cocomittens: Ok, I changed it to just filter for the exact year instead of the range, which resolves the last 2 (though the first one still has 2 release dates). It only misses out on 2 movies (53 instead of 55), which is probably worth the improvement in the accuracy of the data. So yeah, now basically all of them are caused by multiple release dates instead of Powerpuff Girls and Rudolph, not sure what is causing those.

src/data_processing/netflix_to_wikidata.py
cocomittens: Good catch! Done

src/data_processing/netflix_to_wikidata.py
cocomittens: Actually as I put below, it only misses out on 2/100 movies, so updated

src/data_processing/netflix_to_wikidata.py
audiodude: It took a bit of wrangling, but here is the query to filter by US release date: https://w.wiki/B6av

The key is:

```
        ?item p:P577 ?releaseDateStatement .
        ?releaseDateStatement ps:P577 ?releaseDate .
        ?releaseDateStatement pq:P291 wd:Q30 .
```

cocomittens: So apparently only 10% have a US release date
I tried to update it to get either US release date or unspecified which increases it to ~51%

audiodude:
Superseded by work on the `merge-netflix-to-wikidata` branch.

PR #10, opened 2024-08-23
Interaction matrix of user

siddz415:


data_processing/process_data.py
```diff
@@ -0,0 +1,27 @@
...
+        with open(os.path.join(directory_path, filename), 'r') as file:
+            movie_id = int(file.readline().strip().replace(':', ''))
+            movie_idx = movie_id - 1
+            
+            for line in file:
+                user_id, rating, _ = line.strip().split(',')
+                user_id = int(user_id)
+                rating = int(rating)
+                
+                if rating in (4, 5):
```
audiodude, line 20: You can use a pattern to "reverse" your if statements and return early. Instead of:

```
  if condition_a:
     code = blah + foo
     if condition_b:
        some_code = foo + bar
        other_thing()
```

You can write:
```
  if not condition_a:
    return/exit/continue
  code = blah + foo

  if not condition_b:
    return/exit/continue

  some_code = foo + bar
  other_thing()
```

This means you have less levels of indentation and makes your code easier to read. The person reading it doesn't have to think "Okay so X is true and Y is true if we're in this block".

In this case, you would reverse the if condition and `continue` the for loop.

audiodude (changes requested):
This looks good.

I think you might want to think about writing the code that "glues" this all together. Instead of just writing the library functions, what does it looks like to call these functions and produce the final interaction matrix?

data_processing/load_data.py
```diff
@@ -0,0 +1,14 @@
+# This script handles loading the movie titles and ratings data
+import os
+
+def load_movie_titles(filepath):
```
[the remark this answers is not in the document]
siddz415: What do you mean this function is not necessary?

data_processing/process_data.py
```diff
@@ -0,0 +1,27 @@
+# This script handles the processing and creation of the interaction matrix
+
+import numpy as np
+
+def create_interaction_matrix(directory_path, num_users, num_movies, files):
+    interaction_matrix = np.zeros((num_users, num_movies), dtype=np.int8)
+    user_mapper = {}
+    current_user_index = 0
+    
+    for filename in files:
```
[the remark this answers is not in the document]
siddz415: ok, ill refactor it.

data_processing/load_data.py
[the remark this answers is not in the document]
audiodude: Well you don't call it from anywhere in the rest of your code. And it's not clear to me why you would need the titles for calculating the interaction matrix.

data_processing/process_data.py
```diff
@@ -0,0 +1,27 @@
...
+            movie_idx = movie_id - 1
+            
+            for line in file:
+                user_id, rating, _ = line.strip().split(',')
+                user_id = int(user_id)
+                rating = int(rating)
+                
+                if rating in (4, 5):
+                    if user_id not in user_mapper:
+                        user_mapper[user_id] = current_user_index
```
audiodude, line 22: We might want to pickle and save the user_mapper dictionary as well, so we don't have to write a separate function to calculate it again.

data_processing/process_data.py
siddz415: Okay

data_processing/process_data.py
```diff
@@ -17,11 +17,10 @@ def create_interaction_matrix(directory_path, num_users, num_movies, files):
                 user_id = int(user_id)
                 rating = int(rating)
                 
-                if rating in (4, 5):
-                    if user_id not in user_mapper:
+                if rating in (4, 5) and user_id not in user_mapper: # used AND operator instead nested IF
```
audiodude, line 20: I think by combining these if statements you've made the logic incorrect. You want to store the rating in the `interaction_matrix1 even if the user_id is already in the `user_mapper`, right?

So if you see user 1400, and you map it to user 235, you still want to save a second rating for this user if you see it again, right?

audiodude: What I was suggesting before was "returning early". Here's an example with fake function names:

```
def calculate_data(db):
  if db.is_open():
    conn = db.get_conn()
    cursor = conn.get_cursor()
    data = cursor.fetch_data()
    if len(data) > 1:
      update_data(data)
      data.process()
```

Notice that the entire function is indented under an `if` statement. Same thing in the inner if, all of the code is nested underneath it. If we reverse the if condition, we can simply return early and skip the rest of the function. This means that the rest of the function doesn't have to be indented, but more importantly it removes the cognitive overhead when reading the code. Instead of being N levels deep and having to remember "Okay, if I'm here then this and this and this must be true", you can simply reason "okay if there's no db connection we return early" and not have to think about it further.

Here's the same example with "early returns":

```
def calculate_data(db):
  if not db.is_open():
   return

  conn = db.get_conn()
  cursor = conn.get_cursor()
  data = cursor.fetch_data()

  if not len(data) > 1:
    return
   
  update_data(data)
  data.process()
```

audiodude: In your case, since you're in a for loop, it would be early `continue` instead of early `return`.

ivanfslee: Something like this?

```python
            for line in file:
                user_id, rating, _ = line.strip().split(',')
                user_id = int(user_id)
                rating = int(rating)
                
                if rating < 4:
                    continue

                if user_id not in user_mapper:
                    user_mapper[user_id] = current_user_index
                    current_user_index += 1
                    user_idx = user_mapper[user_id]
                    interaction_matrix[user_idx, movie_idx] = rating

                return interaction_matrix


```

audiodude: Esentially yes, though the logic in the second if statement is wrong as you wrote it here, because you need to get the `user_idx` and store `rating` in the `interaction_matrix` whether or not you had to remap the id.

audiodude:
Hey @siddz415 , just wanted to check in if you're able to respond to the comments in this PR? If you're too busy with other stuff that's totally cool. We were thinking we would re-assign the PR if there's not any motion before the meeting on Nov 7. Let us know if you're still interested in working on it.

data_processing/process_data.py
```diff
@@ -1,6 +1,7 @@
 # This script handles the processing and creation of the interaction matrix
 
 import numpy as np
+import os
```
audiodude, line 4: nit: doesn't look like this is used.

jhanley634 (approved):
Using latin-1 encoding in load_movie_titles() seems odd, but perhaps we're loading an ancient file which is not in utf8. It's all commented code in any case, so we should probably just delete those lines before merging to `main`.

The split() on `,` comma suggests that we might prefer to `import csv`.

In lightfm_recommendation.py and main.py we have a lot of code up at module level. Recommend burying it within `def main():` or whatever. That way it will be  safe for some future unit test to `import` it without side effects. The current code, which lacks a `__main__` guard, is innocuous enough. But I'm concerned it will encourage folks to add stuff to this code base which is hard for a "second caller" (such as a unit test) to invoke. Such issues could be addressed in the current PR prior to a merge, or in a subsequent PR.

data_processing/process_data.py
```diff
@@ -0,0 +1,30 @@
+# This script handles the processing and creation of the interaction matrix
+
+import numpy as np
+import os
+
+def create_interaction_matrix(directory_path, num_users, num_movies, files):
+    interaction_matrix = np.zeros((num_users, num_movies), dtype=np.int8)
```
audiodude, line 7: Are you using a coo_matrix here? It looks like a normal one. Even if lightFM allows for a non-sparse matrix, we should use a sparse one to save memory.

audiodude:
> Using latin-1 encoding in load_movie_titles() seems odd, but perhaps we're loading an ancient file which is not in utf8. It's all commented code in any case, so we should probably just delete those lines before merging to `main`.

The file is from 2007, so it's completely possible that it's not UTF-8. It's also possible that it doesn't have any non-ASCII characters anyway, so it might not matter. 

> The split() on `,` comma suggests that we might prefer to `import csv`.

These text files are not CSV, they are only CSV-like. Individual fields can have commas in them. Specifically `movie_titles.txt` often looks like:

```
1023, 1994, Forrest Gump
1024, 2001, Planes, Trains and Automobiles
...
```

> In lightfm_recommendation.py and main.py we have a lot of code up at module level. Recommend burying it within `def main():` or whatever. That way it will be safe for some future unit test to `import` it without side effects. The current code, which lacks a `__main__` guard, is innocuous enough. But I'm concerned it will encourage folks to add stuff to this code base which is hard for a "second caller" (such as a unit test) to invoke. Such issues could be addressed in the current PR prior to a merge, or in a subsequent PR.

I agree. Better to put things in a main function now and figure out how to call it later. We can always rename `main()` to `create_matrix()` or whatever. I think a lot of this comes from the fact that we haven't really discussed how to organize files/modules/libraries/functions, so things are sort of in random places right now.

jhanley634:
This PR is drifting toward being moribund. It is still mergeable at this point, though it proposes adding more than 4 million lines of CSV data, which for `git` is not quite a Best Practice.

It might be helpful to split Data from Code, merge the code, and move on to other issues and feature requests.

audiodude:
Thanks for updating the branch @jhanley634.

We discussed early on in our in-person meetings that our strategy should not be to commit the Netflix prize data. That's the reason we're exploring the other PR that automatically downloads it.

This has already been addressed in this comment: https://github.com/noisebridge/MediaBridge/pull/10#discussion_r1730052989

data_processing/process_data.py
audiodude: On second thought, I'm not sure the user ids are meaningful or useful once we've calculated the matrix.

cocomittens:
I guess my greater question is, what are in these `mv` files exactly?
The first one looks to be a `movie_id: `, followed by a bunch of user ids? (Presumably those that rated that movie?)
Then the second one is user ids and dates?
Where is the rating (presumably on a 1-5 scale?) supposed to come from?

audiodude:
Okay we're clearly working on this at the exact same moment, which probably isn't a great idea. Please pull. I consolidated all of the files and put them in a sane place.

cocomittens:
> Okay we're clearly working on this at the exact same moment, which probably isn't a great idea. Please pull. I consolidated all of the files and put them in a sane place.

Will do! Hopefully didnt create some kind of merge conflict thing, but yeah will make sure to avoid any future interactions

audiodude (approved):

cocomittens (approved):

cocomittens (approved):

audiodude (approved):

PR #13, opened 2024-09-15
Merge netflix to wikidata

cocomittens:
Merges the 2 attempts to map the netflix to wikidata data


src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+    return(wiki_movie_ids, wiki_genres, wiki_directors)
+
+# Calling all functions
+processed_data = []
+
+netflix_file = read_netflix_txt(base_dir + 'netflix_movies.txt')
+
+netflix_csv = base_dir + 'netflix_movies.csv'
+
+wiki_movie_ids_list, wiki_directors_list, wiki_genres_list = wiki_query(netflix_file, user_agent)
```
audiodude, line 123: It looks like you may have reversed some fields here? On line 114 it says you're returning ids, genres and directors. But here you're assigning ids, directors then genres.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+            wiki_movie_ids.append(wiki_feature_info(data, 'item'))
+            wiki_genres.append(wiki_feature_info(data, 'genreLabel'))
+            wiki_directors.append(wiki_feature_info(data, 'directorLabel'))
+        
+    return(wiki_movie_ids, wiki_genres, wiki_directors)
+
+# Calling all functions
+processed_data = []
+
+netflix_file = read_netflix_txt(base_dir + 'netflix_movies.txt')
```
audiodude, line 119: Prefer `os.path.join` over `+`.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+
+netflix_csv = base_dir + 'netflix_movies.csv'
```
audiodude, line 121: Same thing, `os.path.join`.

audiodude, line 121: Usually output is done to an output directory, not the same directory that the input came from.

I picture that the best way to manage the netflix data is to have it at a known location in the repository (`src/data` is fine) where it is git ignored, and then we should also have a path where the output goes.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
+import requests
+import csv
+
+base_dir = os.path.join(os.path.dirname(__file__), '../data/')
```
audiodude, line 4: This is `src/data` right? It probably makes more sense for the data to be at the top level (`MediaBridge/data`) since it's not source code, which `src` implies.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+user_agent = 'Noisebridge MovieBot 0.0.1/Audiodude <audiodude@gmail.com>'
+
+missing_count = 0
+num_rows = 42
+
+# Reading netflix text file
+def read_netflix_txt(txt_file):
+    with open(txt_file, "r", encoding = "ISO-8859-1") as netflix_data:
+        netflix_list = [line.rstrip().split(',', 2) for line in netflix_data.readlines()[0:num_rows]]
+    return(netflix_list)
```
audiodude, line 14: nit: Don't need parentheses for return statements.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+base_dir = os.path.join(os.path.dirname(__file__), '../data/')
+user_agent = 'Noisebridge MovieBot 0.0.1/Audiodude <audiodude@gmail.com>'
+
+missing_count = 0
+num_rows = 42
+
+# Reading netflix text file
+def read_netflix_txt(txt_file):
+    with open(txt_file, "r", encoding = "ISO-8859-1") as netflix_data:
+        netflix_list = [line.rstrip().split(',', 2) for line in netflix_data.readlines()[0:num_rows]]
```
audiodude, line 13: Just be aware here that you're reading all 17000 rows, but throwing away all but `num_rows`. It might be better to do a for loop `for i in range(num_rows):` and then build the list that way.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+
+# Writing netflix csv file
+def create_netflix_csv(csv_name, data_list):   
+    with open(csv_name, 'w') as netflix_csv:
+        csv.writer(netflix_csv).writerows(data_list)
+
+# Extracting movie info from Wiki data
+def wiki_feature_info(data, key):
+    if len(data['results']['bindings']) < 1 or key not in data['results']['bindings'][0]:
+        return 'NA'
```
audiodude, line 24: Is there a reason you're using `'NA'` and not just `None`?

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+    if key == 'genreLabel':
+        return list({d['genreLabel']['value'] for d in data['results']['bindings'] if 'genreLabel' in d})
+    return data['results']['bindings'][0][key]['value'].split('/')[-1]
+
+# Adding movie info to netflix csv file
+def add_movie_info_to_csv(data_csv, movies, genres, directors):
```
audiodude, line 30: This doesn't look like it's used anywhere. And I would be worried if it was, because it is very inefficient to read a CSV file and change one row and write it back out.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+        csv_writer = csv.writer(csv_file)
+        csv_writer.writerows(data_list)  
+
+# Getting list of movie IDs, genre IDs, and director IDs from request
+def wiki_query(data_csv, user_agent):
+    wiki_movie_ids = []
+    wiki_genres = []
+    wiki_directors = []
+    
+    # csv_reader = list(csv.reader(open(data_csv)))
```
audiodude, line 49: nit: remove commented out code.

src/data_processing/wiki_to_netflix.py
audiodude, line 119: nit: Probably shouldn't be called `netflix_file` because it's not a file, it's a list of lists right?

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+    
+    for row in data_csv:
+        if row[1] != "NULL":
```
audiodude, line 52: Same thing as on the other PR, invert your ifs. Instead of an if, with 100 lines of code indented underneath it, reverse the if condition and exit early. So:

```
if row[1] == 'NULL':
  continue
```

And then outdent the rest of the stuff in the for loop.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,136 @@
...
+processed_data = []
+
+netflix_file = read_netflix_txt(base_dir + 'netflix_movies.txt')
+
+netflix_csv = base_dir + 'netflix_movies.csv'
+
+wiki_movie_ids_list, wiki_directors_list, wiki_genres_list = wiki_query(netflix_file, user_agent)
+
+for index, row in enumerate(netflix_file):
+    title, year, netflix_id = row[0], int(row[1]), row[2]
```
audiodude, line 126: Nit: You can just put `row` on the right side, and as long as the `row` list only has three elements, Python will destructure it for you.

```
title, year, netflix_id = row
```

audiodude: In fact, you can even put the tuple right in your for loop:

```
for index, (title, year, netflix_id) in enumerate(netflix_file):
    ...
```

audiodude (changes requested):
This looks like it works and it doesn't have any major issues, good job.

I wonder if there could be an improvement in the general layout of this code though. The `wiki_query` method seems to iterate over all of the rows in the `netflix_data`. Would it make more sense to iterate "outside" that query, and do something like:

```
for title, year, netflix_id in netflix_data:
  wikidata_id, genres, directors = wiki_query(title, year)
  processed_data.append(netflix_id, wikidata_id, title, year, genres, directors)
```

This is even more appealing since we're going to be putting everything in Mongo. Rather than build up these 17000 item lists of everything, you could process the data one line at a time and store it in Mongo after you'e done the query, then forget about it.

In fact, it would make sense to read the Netflix file line by line and yield it's fields as a generator than to read it all at once.

src/data_processing/wiki_to_netflix.py
cocomittens: It is kind of a holdover from the previous method of reading and writing to the same csv and adding NA to the missing columns. Which was refactored to creating the csv and adding only the found rows at the end.

src/data_processing/wiki_to_netflix.py
cocomittens: Yeah, this is also a holdover from before the refactor which isn't currently being used. It has been deleted

src/data_processing/wiki_to_netflix.py
cocomittens: Yeah, I'm using the `src/data` folder which was already there (but empty). I could change it/ add a new one, but wasn't sure if there was a reason it is where it is currently.

src/data_processing/wiki_to_netflix.py
cocomittens: Fixed

src/data_processing/wiki_to_netflix.py
cocomittens: Fixed

src/data_processing/wiki_to_netflix.py
cocomittens: Fixed

src/data_processing/wiki_to_netflix.py
cocomittens: How about `netflix_data`?

src/data_processing/wiki_to_netflix.py
cocomittens: Fixed

src/data_processing/wiki_to_netflix.py
cocomittens: Yeah I think it got mixed up. Should be fixed

cocomittens:
By the way, the error when running all the movies is -
`TimeoutError: [Errno 60] Operation timed out`

src/data_processing/wiki_to_netflix.py
audiodude: Yeah I don't know where that path came from, but I think it makes more sense for the `data` directory to be at the top level.

src/data_processing/wiki_to_netflix.py
audiodude: Do you think we should change it to `None`?

audiodude (changes requested):

src/data_processing/wiki_to_netflix.py
audiodude: Yes that's good.

.gitignore
```diff
@@ -0,0 +1,2 @@
+src/data/netflix_movies.txt
```
audiodude, line 1: Remember to update this file when we change the input/output directory locations.

I recommend

```
data/
out/
```

and just ignore those two directories

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,134 @@
...
+def wiki_query(data_csv, user_agent):
+    wiki_movie_ids = []
+    wiki_genres = []
+    wiki_directors = []
+        
+    for row in data_csv:
+        if row[1] == "NULL":
+            continue
+
+        SPARQL = '''
```
audiodude, line 45: Maybe move the triple quoted string up to the top level and leave it as the constant `SPARQL` then do something like:

```
query = SPARQL % {'Title': row[2], 'Year': int(row[1])}
```

That would make this function a lot easier to read.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,134 @@
...
+    for index, row in enumerate(netflix_file):
+        netflix_id, year, title = row[0], int(row[1]), row[2]
+        if wiki_movie_ids_list[index] == 'NA':
+            missing_count += 1
+        else:
+            processed_data.append([netflix_id, wiki_movie_ids_list[index], title, year, wiki_genres_list[index], wiki_directors_list[index]])
+
+    create_netflix_csv(netflix_csv, processed_data)
+
+    print('missing:', missing_count, '(', missing_count / num_rows * 100, '%)')
```
audiodude, line 129: You can do this all with `f''` strings if you want. You can put arithmetic expressions in `f''` strings:

```
print(f'found: {num_rows - missing_count}')

**Review by audiodude — COMMENTED** [review]
A few more things before I forget.

File: src/data_processing/wiki_to_netflix.py [hunk 15]
```diff
@@ -0,0 +1,133 @@
...
+import os
+import math
+
+base_dir = os.path.join(os.path.dirname(__file__), '../data/')
+user_agent = 'Noisebridge MovieBot 0.0.1/Audiodude <audiodude@gmail.com>'
+
+# Reading netflix text file
+def read_netflix_txt(txt_file, test):
+    netflix_list = []
+    num_rows = 100 if test else math.inf
```
audiodude, line 12: It's a bit messy to se num_rows to infinity. It would be better to just set it to `None` if `test == True` and check for that in the condition below.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,133 @@
...
+
+    with open(txt_file, "r", encoding = "ISO-8859-1") as netflix_data:
+        for i, line in enumerate(netflix_data):
+            if i >= num_rows:
```
audiodude, line 16:
```
if num_rows is not None and i >= num_rows:
```

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,133 @@
...
+        wiki_directors.append(wiki_feature_info(data, 'directorLabel'))
+    
+    return(wiki_movie_ids, wiki_genres, wiki_directors)
+
+# Calling all functions
+def process_data(test=False):
+    missing_count = 0
+    processed_data = []
+
+    netflix_file = read_netflix_txt(os.path.join(base_dir, 'netflix_movies.txt'), test)
```
audiodude, line 113: Maybe we should leave this as `movie_data.txt` so that we as developers can simply extract the netflix data set to `/data` and not have to rename the file?

src/db/insert_data.py
```diff
@@ -1 +1,35 @@
-# Script to insert movies and interactions into MongoDB
\ No newline at end of file
+# Script to insert movies and interactions into MongoDB
```
audiodude, line 1: This might be best in the next PR....

src/data_processing/wiki_to_netflix.py
cocomittens: That is fair tbh. I didn't realize that you could compare integers to `None` in python, and that it apparently will work as intended. I agree the inline if/ else statement is a bit messy/ lazy as well. This was kinda just like minor tech debt that was there because it technically worked, but yeah should be fixed now.

src/data_processing/wiki_to_netflix.py
cocomittens: Yeah that makes sense. I forgot what the original name was tbh. Fixed

src/data_processing/wiki_to_netflix.py
cocomittens: I did that and refactored the query construction to be in its own function. Should be a bit more solid.

src/data_processing/wiki_to_netflix.py
audiodude: See comment below for a suggested way to write the if, if num_rows is `None`

src/data_processing/wiki_to_netflix.py
audiodude: I dont' see those changes, did you push them?

audiodude (commented):
I started reviewing this again, but it seems to be missing changes you describe in your comments. Can you double check that you pushed to GitHub?

cocomittens:
> I started reviewing this again, but it seems to be missing changes you describe in your comments. Can you double check that you pushed to GitHub?

Hmm I feel like they should be there? These should be them
Screen Shot 2024-09-29 at 3 18 08 PM

cocomittens:
> > I started reviewing this again, but it seems to be missing changes you describe in your comments. Can you double check that you pushed to GitHub?
> 
> Hmm I feel like they should be there? These should be them Screen Shot 2024-09-29 at 3 18 08 PM

Oh wait, I realize there may have been some complications with me accidentally pushing this with my other account. RIP green squares.

audiodude:
Hi @cocomittens @smaysenhalder 

This PR still has a lot of comments on it. Because of lots of commits and changing of the file, many of the comments are marked as "Outdated", but that simply means they were made on an old version of the code. The questions and suggestions still apply.

Any comment on this page, the "conversation" page, that isn't marked as resolved still needs to be either 1) fixed in the code or 2) commented on to explain why it shouldn't be fixed or why it needs to be discussed further.

Unfortunately, you won't see the "outdated" comments when you look at the code itself.

Thanks!

src/data_processing/wiki_to_netflix.py
audiodude: Okay I see it now thanks!

src/data_processing/wiki_to_netflix.py
audiodude: Sorry, my mistake, the file is called `movie_titles.txt`.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,141 @@
...
+        wiki_directors.append(wiki_feature_info(data, 'directorLabel'))
+    
+    return(wiki_movie_ids, wiki_genres, wiki_directors)
+
+# Calling all functions
+def process_data(test=False):
+    missing_count = 0
+    processed_data = []
+
+    netflix_file = read_netflix_txt(os.path.join(base_dir, 'movie_data.txt'), test)
```
audiodude, line 121:
```suggestion
    netflix_file = read_netflix_txt(os.path.join(base_dir, 'movie_titles.txt'), test)
```

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,141 @@
...
+        netflix_id, year, title = row[0], int(row[1]), row[2]
+        if wiki_movie_ids_list[index] == 'NULL':
+            missing_count += 1
+        processed_data.append([netflix_id, wiki_movie_ids_list[index], title, year, wiki_genres_list[index], wiki_directors_list[index]])
+
+    create_netflix_csv(netflix_csv, processed_data)
+
+    print('missing:', missing_count, '(', missing_count / num_rows * 100, '%)')
+    print('found:', num_rows - missing_count, '(', (num_rows - missing_count) / num_rows * 100, '%)')
+    print('total:', num_rows)
```
audiodude, line 138:
```suggestion
    print(f'missing:  {missing_count} ({missing_count / num_rows * 100}%)')
    print(f'found: {num_rows - missing_count} ({num_rows - missing_count) / num_rows * 100}%)')
    print(f'total: {num_rows}')
```

src/data_processing/wiki_to_netflix.py
audiodude: See suggestion below

src/db/insert_data.py
audiodude: Though honestly there's like 83 thousand commits in this PR now, and it might not be worth it to try and remove this. Let's just try not to add any more features before we check this in.

src/data_processing/wiki_to_netflix.py
smaysenhalder: I can update this using the first solution; Kind of curious if you know why it doesn't raise an error for year / how it knows to treat row[1] as int? (because the initial solution explicitly converts str year to int year)

src/data_processing/wiki_to_netflix.py
smaysenhalder: Will remove parenthesis throughout for return statements

src/data_processing/wiki_to_netflix.py
audiodude: You're right. Technically my suggestion doesn't actually map one to one with what is currently there. If you destructure the row, you will get 3 strings for title, year and netflix_id.

However I ignored the cast to `int` because it's unnecessary. Yes the year is an integer. But it's fine leaving it as a str because it's just being used to format the SPARQL query. So if it goes into the format as an int, it gets converted back to str and placed in the query as a string anyway.

Even when it is output, it is being written to a CSV, so that's also a string.

src/data_processing/wiki_to_netflix.py
smaysenhalder: I changed it to "NULL" so that Mongo stores unmatched movies with a wikidata_id (and other feature info) of NULL (based on the comment in the other issue); Curious if/why None is typically preferred?

src/data_processing/wiki_to_netflix.py
smaysenhalder: Updated to movie_titles.txt

src/data_processing/wiki_to_netflix.py
audiodude: Just to be clear, the way I'm picturing this is that someone who wants to use this repo will download the netflix data, and extract it to `MediaBridge/data` and then we basically treat that directory as "read only". We don't write any files there, and we don't modify anything there. We don't touch it. That way, once the data is "installed" no one has to think about it ever again.

src/data_processing/wiki_to_netflix.py
smaysenhalder: Gotcha, thanks!

src/data_processing/wiki_to_netflix.py
audiodude: For Mongo, it's okay to simply leave out fields that aren't known or required. So you can have one movie:

```
{
  "title": "Jurassic Park",
  "year": "1996",  # Note, this is when it actually makes sense to have the year as an int!
  "wikidata_id": "Q56789",
  "genres": ["Adventure", "Family", "Action"]
  "director": "Stephen Spielberg"
}
```

and another movie:

```
{
  "title": "Wacky Adventures With Fruits",
  "year": "1999",
}
```

And they can both exist in the same Mongo collection. When reading the data, we will have to do a check like `if 'wikidata_id' in movie_data:`, but this is no more work then doing `if movie['wikidata_id'] == 'NULL'` and the database is cleaner.

audiodude: There is a big difference between the string `"NULL"` and the value `None`.

The first is a string, obviously. It doesn't have any special meaning. It might as well be "APPLE".

`None` signifies the absence of a value. Think of fruits. If we don't know the price of a fruit, but we know we have the fruit, we can use `None` to signify that the value is unknown/not present:

```
fruits = {'apple": 1.29, 'banana': None, 'kiwi": 1.89}
```

We need something to put after the banana key, but we can't put 0 (that would imply that bananas are free). We could maybe use what's called a "sentinel value" of -1 or something, but then someone who is calculating the price of all the fruits might not realize that the sentinel is there:

```
total_price = sum(fruits.values())
```

On the other hand, if the value is `None`, the above code will raise an exception. I sometimes find it easiest to think of numbers/integers. If the bank balances is 1, you have one dollar. If it's -1, you owe one dollar. If it's 0, you have no money. How do we specify that your account is closed and there is "no" balance? We use `None`.

The other thing that may be confusing is that in other languages like SQL, the name `NULL` is used (no quotes!) for the absence of a value such as:

```
...WHERE wikidata_id IS NULL
```

In Ruby, this is called `nil`. It's all the same thing: there is no value (very zen)

src/data_processing/wiki_to_netflix.py
audiodude: Just a quick update, as said in another comment `src/data` is probably not good because the data is not technically source code. Probably should be `MediaBridge/data`.

audiodude (changes requested):
`MediaBridge` is the "root" of this project. So if you have code at `/users/me/Documents/code`, the `/users/me/Documents/code/MediaBridge/` directory would contain all of the files that you see on the "Code" tab of the repo.

So it follows that there's no `MediaBridge` folder inside `MediaBridge`. This .gitignore asks git to ignore the `/users/me/Documents/code/MediaBridge/MediaBridge/data` directory. It should just be:

```
data/
out/
```

smaysenhalder:
> `MediaBridge` is the "root" of this project. So if you have code at `/users/me/Documents/code`, the `/users/me/Documents/code/MediaBridge/` directory would contain all of the files that you see on the "Code" tab of the repo.
> 
> So it follows that there's no `MediaBridge` folder inside `MediaBridge`. This .gitignore asks git to ignore the `/users/me/Documents/code/MediaBridge/MediaBridge/data` directory. It should just be:
> 
> ```
> data/
> out/
> ```

.gitignore
smaysenhalder: Updated the .gitignore file to remove  "MediaBridge" from the directory path so that it's no longer treating it as a folder that stores "data" and out" but as the root of the repo.

.gitignore
```diff
@@ -0,0 +1,3 @@
+/data
+/out
```
audiodude, line 2: This works but it's a little non-standard. The more standard way would be:

```
data
out
```

or

```
data/
out/
```

The trailing slash is only necessary to indicate to readers of this file that they are directories.

data/.gitignore
audiodude, line 1: The `.gitignore` files cascade. What this means is that if you ignore `.DS_store` in `MediaBridge/` (aka the root directory), it will be ignored in all subdirectories. If instead you had a `.gitignore` in `MediaBridge/foo/bar` with `.DS_Store` listed, it would be ignored only in `MediaBridge/foo/bar/` and its subdirectories, not in `MediaBridge/foo/` or `MediaBridge`.

What I'm saying is that you don't need an extra, empty `.gitignore` file in this directory.

However, as you may have noticed, you can't "track" (as git calls it) directories themselves, only files. So if a directory is empty, it can't be added to git. Developers have come up with the convention of adding an empty `.gitkeep` file, which has no technical function, just to indicate that the directory should be "kept" in the repo (as opposed to being "pruned").

In reality, though, you don't need a `.gitkeep` file either. You don't need these directories in the repo, especially if they're going to be in the `.gitignore` file. The point of having them in that file is so they *don't* get added to the repo.

cocomittens:
Wait why did the PR get closed?

audiodude:
> Wait why did the PR get closed?

Just by mistake, thanks for reopening.

data/.gitignore
smaysenhalder: Got it, thanks for explaining! I'll go ahead and remove them and then resolve comment

.gitignore
smaysenhalder: Updated

src/data_processing/wiki_to_netflix.py
audiodude: Path should be `../../data`

audiodude: Also `base_dir` should be called `data_dir` or something.

src/data_processing/wiki_to_netflix.py
audiodude: Prob ably need a new variable for the output directory, `out_dir`.

And probably should use `os.path.join` instead of string concatenation.

audiodude (commented):
Sorry, I forgot to send my review comments last night, here they are.

As I mentioned, there's a lot of random stuff in this commit, so it's up to you if you want to fix all the things I mentioned or just revert the commit and start fresh.

audiodude (commented):
Sorry, can you explain the reasoning behind this change?

cocomittens:
> Sorry, can you explain the reasoning behind this change?
> 
> EDIT: The "Refactor file structure" one.

Well basically,
I was initially addressing the above comment that referenced what some would call a refactor, while others would call it something else (I am unsure what exactly)
I was feeling the urge to simply tear it all down and was like yolo, this is an anarchist code base for a minute, where there are no hierarchical structures. At least until I figure out where to put them. Which may possibly be exactly where they came from, or not.

Then I felt the need to quicksave my progress on this restructuring or some such in case my device were to spontaneously combust, decided to commit this change out of habit because it felt like progress. Before I fully thought through this decision, I had already typed and submitted the `git push` command which often follows such a commit. I then got distracted with this other project I have to do and intended to return to it after that. However I will fix this, don't worry

That is my summary of what prompted this change.

cocomittens:
> > Sorry, can you explain the reasoning behind this change?
> > EDIT: The "Refactor file structure" one.
> 
> Well basically, I was initially addressing the above comment that referenced what some would call a refactor, while others would call it something else (I am unsure what exactly) I was feeling the urge to simply tear it all down and was like yolo, this is an anarchist code base for a minute, where there are no hierarchical structures. At least until I figure out where to put them. Which may possibly be exactly where they came from, or not.
> 
> Then I felt the need to quicksave my progress on this restructuring or some such in case my device were to spontaneously combust, decided to commit this change out of habit because it felt like progress. Before I fully thought through this decision, I had already typed and submitted the `git push` command which often follows such a commit. I then got distracted with this other project I have to do and intended to return to it after that. However I will fix this, don't worry
> 
> That is my summary of what prompted this change.

Okay, order has been restored to the file structure & I shall continue to address any remaining bugs/ concerns

audiodude:
> > Sorry, can you explain the reasoning behind this change?
> > EDIT: The "Refactor file structure" one.
> 
> Well basically,

That was a very amusing description thank you! 😂

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,137 @@
...
+
+# Calling all functions
+def process_data(test=False):
+    missing_count = 0
+    processed_data = []
+
+    netflix_file = read_netflix_txt(os.path.join(base_dir, 'movie_titles.txt'), test)
+    num_rows = len(netflix_file)
+
+    netflix_csv = os.path.join(base_dir, 'movie_titles.csv')
```
audiodude, line 119: It looks like we're still outputting a CSV to the data directory (because `base_dir` is actually `data_dir` and refers to `../../data`, see other comment).

As seen in the `.gitignore` file, let's use a top level `out` directory for the output: `../../out`. You should probably rename `base_dir` while you're at it to avoid confusion.

src/data_processing/wiki_to_netflix.py
```diff
@@ -0,0 +1,137 @@
...
+        wiki_directors.append(wiki_feature_info(data, 'directorLabel'))
+    
+    return wiki_movie_ids, wiki_genres, wiki_directors
+
+# Calling all functions
+def process_data(test=False):
+    missing_count = 0
+    processed_data = []
+
+    netflix_file = read_netflix_txt(os.path.join(base_dir, 'movie_titles.txt'), test)
```
audiodude, line 116: This still says `netflix_file`. We were going to rename it to `netflix_data` right?

src/data_processing/wiki_to_netflix.py
audiodude: See other comment.

src/data_processing/wiki_to_netflix.py
audiodude: See other comment.

audiodude (commented):
I think we're almost there!

I've resolved any comments that have already been addressed, and tried to create new comments that are not "outdated" to point out issues that are still there.

Also, remind me to discuss this at our next meeting, but I feel the best workflow for code reviews is:

1. Reviewer leaves a comment
2. Comment is discussed (optional)
3. Author replies to comment with "Done" or "Done, updated the variable to blah blah blah"
4. Reviewer marks comment as resolved.

That way the reviewer ultimately decides that the feedback has been addressed, but step 3 ensures that the reviewer knows which comments are ready for another look and doesn't have to go hunting through the code.

src/data_processing/wiki_to_netflix.py
smaysenhalder: Yep, went ahead and updated to netflix_data :))

src/data_processing/wiki_to_netflix.py
smaysenhalder: Just updated to have two separate directories for the data folder (data_dir) and out folder (out_dir) for review

src/db/insert_data.py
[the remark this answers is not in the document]
smaysenhalder: Will go ahead and remove this file right now and have it included in the next PR

src/data_processing/wiki_to_netflix.py
audiodude: Probably not an actual issue because it's only 17000 rows. In "production" we are reading all the rows anyway. Resolving.

audiodude:
Okay all comments are resolved! 🎉 

Is there anything else anyone wants to do or can I merge?

audiodude (approved):

cocomittens:
> Okay all comments are resolved! 🎉
> 
> Is there anything else anyone wants to do or can I merge?

MERGE IT!!!! :smiley: can deal with whatever else in future PRs

cocomittens:
Its so green and not red, SO ready to be merged and rehomed so it can live on main now 👀

audiodude:
> Its so green and not red, SO ready to be merged and rehomed so it can live on main now 👀

@cocomittens would you like to do the honors? You're a collaborator on the project so you should be able to click that big, shiny green button!

Issue #12, opened 2024-09-13
Figure out why we can't run the netflix_to_wikidata script on all the movies


audiodude:
It seems to be crashing part of the way through. Let's post the stack trace in this bug and see if we can figure it out.

audiodude:
Looks like the network requests are just a bit flaky and eventually one gets stuck and times out. Maybe we're getting rate limited (we should look into that).

https://github.com/noisebridge/MediaBridge/pull/13#issuecomment-2354287441

audiodude:
https://www.wikidata.org/wiki/Wikidata:REST_API#Rate_limits

audiodude:
Okay, I ran the code with the version in #17, with the exponential backoff and with `tqdm` showing a progress bar. It failed 27% of the way through, but with a different error message than the one we got before:

```
27%|█████████████████▉                                                | 4824/17770 [28:09<1:15:33,  2.86it/s]
Traceback (most recent call last):
  File "/home/tmoney/code/MediaBridge/src/data_processing/wiki_to_netflix.py", line 152, in <module>
    process_data(False)
  File "/home/tmoney/code/MediaBridge/src/data_processing/wiki_to_netflix.py", line 137, in process_data
    wiki_movie_ids_list, wiki_genres_list, wiki_directors_list = wiki_query(netflix_file, user_agent)
                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/tmoney/code/MediaBridge/src/data_processing/wiki_to_netflix.py", line 118, in wiki_query
    response.raise_for_status()
  File "/home/tmoney/.local/share/virtualenvs/MediaBridge-QSS_14Zx/lib/python3.12/site-packages/requests/models.py", line 953, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://query.wikidata.org/sparql
```

A 400 error means that there was something wrong with the request, with what we sent Wikidata. My guess is that there is a movie name like "Face/Off" that has a slash in the name or something, and it's not getting properly escaped, which would make the SPARQL invalid. Picture a movie like `Joe's "Magical" Adventure`. When put into the SPARQL it would be:

```
                                mwapi:search "%(Title)s" ;
```

which would turn into

```
                                mwapi:search "Joe's "Magical" Adventure" ;
```

[...]

Issue #15, opened 2024-09-27
Think about adding tests


audiodude:
(self explanatory)

jhanley634:
I feel the best thing you can do is arrange for `$ make test` to run `pytest --cov --cov-report=term-missing`. (It's also possible to run `coverage erase` and friends, but that's a bit more involved.)

If a feature-branch author sees they added a bunch of code which is called as being "not covered", then they _might_ be sufficiently embarrassed that they author the corresponding test. Or not, whatever. But hope springs eternal.

audiodude:
Possbily, but first we have to have tests at all! This issue is to track that.

audiodude:
@cocomittens 

Let's do the structure we talked about (`foo.py`, `foo_test.py`, `foo_testdata.py`). It's fine to have a PR with just one simple test in it. Please document in the README how to run the tests (probably `pipenv run pytest`).

audiodude:
Tests have been added and CI turned on.

PR #17, opened 2024-09-27
Try exponential backoff to avoid read errors

audiodude:
Possible fix for #12


jhanley634:
This is perfectly nice code and could be merged to `main` as-is.

I will note in passing that
1. The indent level gets slightly deep, and a [@tenacity](https://pypi.org/project/tenacity) decorator would be a great, already-tested solution for this.
2. PR #17 claims to fix a bug, change a behavior, but it does not include an automated unit test which verifies such a claim. Prefer to "write a failing test" for bugfixes, and then the PR code turns a Red test into Green.

nit: The `except requests.exceptions.Timeout:` is tediously long. Prefer to put the verbiage into the imports section, which everyone just glances over, so we have a nice concise `except Timeout:` clause.

Oh, and kudos on the `.raise_for_status()`, that is _always_ a good plan.

audiodude:
> This is perfectly nice code and could be merged to `main` as-is.

Thanks for the review! 


> I will note in passing that
> 
> 1. The indent level gets slightly deep, and a [@tenacity](https://pypi.org/project/tenacity) decorator would be a great, already-tested solution for this.

I looked up `tenacity`, and it seems like it would add an unnecessary layer of abstraction, making the code harder to reason about. It would also necessitate restructuring the code. Additionally it seems to require learning its DSL for things like `wait=wait_fixed(3) + wait_random(0, 2)`. It is not immediately obvious what this code does.

> 2. PR [Try exponential backoff to avoid read errors #17](https://github.com/noisebridge/MediaBridge/pull/17) claims to fix a bug, change a behavior, but it does not include an automated unit test which verifies such a claim. Prefer to "write a failing test" for bugfixes, and then the PR code turns a Red test into Green. And yeah, I understand that will tend to lead you down the Mocking path so you can deliberately inject failures on demand. It's a little more ceremony. But hey, this is supposed to be a learning exercise, right, it's all good.

Yes automated tests would be nice, see #15.

> nit: The `except requests.exceptions.Timeout:` is tediously long. Prefer to put the verbiage into the imports section, which everyone just glances over, so we have a nice concise `except Timeout:` clause.

Personally I prefer the verbosity, because it makes it clear where the exception is sourced from. Something like `except Timeout` could possibly confuse readers of the code as to if `Timeout` is a built-in exception, or maybe it's coming from `urllib3` or even `socket` or something.

> Oh, and kudos on the `.raise_for_status()`, that is _always_ a good plan.

Thanks!

cocomittens:
There were some indentation issues, but I fixed them

cocomittens:
Wait, might have broken it actually. I'll revert but yeah was giving an error so will figure that out.

cocomittens:
Actually I think it should be fixed now
Sry to commit on your pr, just wanted to fix that real quick so can test it now

cocomittens:
Looks solid, tested it and it works. Don't have time to look at it more closely before the meeting, but I feel like it should be fine to merge even if it's refactored a bit later.

cocomittens (approved):

jhanley634:
I am trying to click Approve, but I think this will wind up instead being a Comment which does not affect whether merging is allowed.

Here's a trivial nit, in mediabridge/data_processing/wiki_to_netflix.py:
We're not following the [PEP 8](https://peps.python.org/pep-0008/#imports) advice on organizing `import`s into three sections.
It is an aid to the reader, makes it easier to what is familiar Batteries Included (standard docs) and what is documented over on pypi + project homepage.
There's no need to waste brain cells on such minutiae -- let [isort](https://pypi.org/project/isort) sweat the details. I recommend just merging it as-is, and let a subsequent PR do a routine `$ isort .`

Anyway, the tqdm progress bar is definitely a Good Thing.

Different nit: mathematicians would typically prefer to phrase `2 ** tries * 5` as `5 * 2 ** tries`.

----

Thank you for the `__main__` guard.

Clearly the `test` assignment works. But the fact we felt a (helpful!) comment was needed suggests that maybe we're working too hard. Be lazy! Let [typer](https://pypi.org/project/typer) sweat the details about cracking argv. Add a `prod: bool` parameter to the `process_data()` signature. And no, you certainly don't need to do that in _this_ PR, you might defer that to a subsequent PR.

I have some remarks that are clearly out-of-scope for the current PR. We have some very nice function preamble comments:

- 15:# Reading netflix text file
- 27:# Writing netflix csv file
- 32:# Extracting movie info from Wiki data
- 40:# Formatting SPARQL query for Wiki data
- 89:# Getting list of movie IDs, genre IDs, and director IDs from request
- 132:# Calling all functions

But they should all be """docstrings""", explaining the [contract](https://en.wikipedia.org/wiki/Design_by_contract) the function is bound by. Consider addressing that in a subsequent PR.

audiodude:
Thanks for the review @jhanley634! If you'd like, I can add you to the contributors of the project so that you can actually approve when you do a code review. Or you can join [this group](https://github.com/orgs/noisebridge/teams/pythonprojectmembers) and get write access. I don't know how to add people to that group or how to manage it, so I've just been adding people to the project directly.

Anyways, I've created issues for the major points of your feedback, thanks again!

jhanley634:
Yes, I think that makes sense. Please add me to the repo. (The group URL you mentioned is apparently private, as it 404’d for me.)

Personal philosophy: I feel the PR process should not slow things down, should not be an obstacle. If it has been 24 hours without comment, it’s appropriate to consider the PR “auto approved”, and just merge it down. I view each PR as a teaching opportunity.

    Cheers,
    JH

audiodude:
Okay I've added you!

I agree in principle that we shouldn't have to wait indefinitely for someone to merge our PRs. And I agree also that this isn't some mission critical code. There aren't even stakeholders, let alone customers, so merging bad code is probably No Big Deal.

All that said, 24 hours is not long enough. Maybe 72? Everyone has day jobs and this is a side project, and I would much prefer that code sits around for a few days and "slows things down" than to have lots of unreviewed code being thrown at the project all at once.

And with _that_ said, I will remind everyone in this project that they can and should review each other's code. I should not be a bottleneck here.

skyfenton:
Def dropped the ball a bit, I procrastinated reviewing this while prepping for an interview, then ran out of time before the vacation I just came back from. Will be giving more attention to assigned PRs!

audiodude:
sg, no worries!

PR #21, opened 2024-10-10
Allow main file to import data into mongo

smaysenhalder:


audiodude:
So I think this code looks pretty good, I haven't gotten a chance to review it in depth. However I think you're doing "too much" in one PR. Let's try to keep the PRs small and focused. There should've been:

1. PR to put data/output directory in a constants file
2. PR to escape quotes
3. PR to include the Mongo code and even maybe a separate PR to integrate it into the main file.

I don't think it's worth it to break it up at this point, but just something to think about in the future. In general it should be "I have this task, so I'll create a branch/PR for it", not "I have this branch/PR, so whatever tasks I have will go there".

Thanks!

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -4,12 +4,15 @@
 import time
 from tqdm import tqdm
 import sys
+from db.insert_data import insert_into_mongo
+from constants import *
```
audiodude, line 8: The use of `import *` is an anti-pattern. It results in a bunch of random 'magical' variables being in your script, with readers of the script having no way of knowing where they came from. For instance, if someone sees `DATA_DIR` below and looks at the imports, they would have to assume it comes from this import. But what if you had multiple `import *` statements? The variable could be coming from any of them.

Prefer `from constants import DATA_DIR, OUT_DIR` or `import constants` and then using `constants.DATA_DIR` below.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -4,12 +4,15 @@
...
 
 class WikidataServiceTimeoutException(Exception):
     pass
 
-data_dir = os.path.join(os.path.dirname(__file__), '../../data')
-out_dir = os.path.join(os.path.dirname(__file__), '../../out')
+# data_dir = os.path.join(os.path.dirname(__file__), '../../data')
```
audiodude, line 13: nit: no need to leave commented code, unless it will be used in a future PR.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -95,6 +98,9 @@ def wiki_query(data_csv, user_agent):
     for row in tqdm(data_csv):
       if row[1] is None:
             continue
+          
+      if '"' in row[2]:
+            row[2] = row[2].replace('"','\\"')
```
audiodude, line 103: Just wanted to double check that you confirmed that `\"` is the proper way to escape quotes in SPARQL. I think I spitballed a bit and used that as an example, but I have no idea if it's correct.

src/constants.py
audiodude, line 1: The `src` directory no longer exists, and should never exist again. This should be `mediabridge/constants.py`.

src/main.py
audiodude, line 1: Same thing here, move to `mediabridge/main.py`

src/main.py
```diff
@@ -0,0 +1,8 @@
+# Main entry point to run the recommendation pipeline
+
+from data_processing.wiki_to_netflix import * 
+
+if __name__ == '__main__':
+    test = len(sys.argv) < 2 or sys.argv[1] != '--prod'
```
audiodude, line 6: nit: Looks like we lost the comment here, can we try to re-add it?

src/main.py
```diff
@@ -0,0 +1,8 @@
+# Main entry point to run the recommendation pipeline
```
audiodude, line 1: nit: this comment is nonsense, we should probably just remove it.

audiodude (changes requested):
Thanks for making all these changes. I hope I wasn't too harsh about the "one change per PR" thing, just trying to give guidance.

mediabridge/data_processing/wiki_to_netflix.py
smaysenhalder: Have removed them

jhanley634 (approved):
Please do _not_ merge down the "from constants import *" line. It needs to mention specific symbols, not `*`, as an aid to the reader. We read code much more than we write it.

It looks like PR #17 https://github.com/noisebridge/MediaBridge/pull/17 should probably be merged down to `main` before this PR is.

mediabridge/data_processing/wiki_to_netflix.py
smaysenhalder: I went the route of replacing any double quotes with '\\"' in the titles that include doubles quotes (which escapes quotes) before these titles reach the SPARQL query portion. So instead of having the query deal with escaping the quotes, the titles are reformatted beforehand to escape this. I created a sample .txt file with titles that all have double quotes to test it out and it worked, but let me know if you're fine with me using this logic & going about it this way.

mediabridge/data_processing/wiki_to_netflix.py
smaysenhalder: That's a good point, just updated :)

src/main.py
smaysenhalder: Removed it

src/constants.py
smaysenhalder: Sorry about that Travis! I had relocated these two to the main mediabridge directory at some point in another branch so it didn't translate to this one as this one was created much earlier. My other branch that adds logging info might also have the old src structure. Wondering if maybe I should just delete that branch and then create a new branch off of the main file so it has the most recent structure since I'd only be adding a few lines of code and want to make sure the structure is the most up-to-date.

src/main.py
smaysenhalder: Updated (further explanation in comment above)

src/constants.py
audiodude: If you're only adding a few lines of code, yes it probably makes sense to branch off of main and redo the changes. Sorry about that!

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Yes that's fine. Testing it directly is totally acceptable. Good work.

main.py
audiodude, line 1: I can't really tell for some reason, but it looks like this file is in the root directory. It should be in mediabridge (not MediaBridge the main directory that the project is in).

If you have a path on your computer like `/Users/me/Documents/code/MediaBridge/`, that is the "root" of the project. If you notice, everything on the main "code" page of GitHub, like `.gitignore` and `README.md` are in that "root" directory. So you have `/Users/me/Documents/code/MediaBridge/README.md`.

It appears to me that this commit adds a file at `/Users/me/Documents/code/MediaBridge/main.py` where it should actually be `/Users/me/Documents/code/MediaBridge/mediabridge/main.py`

audiodude: Same thing with constants.py

audiodude (commented):
Great job on these changes, we were able to resolve a lot of the conversations. Still a few things to do though.

main.py
smaysenhalder: I relocated these two to the mediabridge file. There was another file under there called "main" that had the following code:

jhanley634:
This is shaping up nicely, looks good.

`from ... import *` is not a Best Practice. Prefer to replace "*" with "insert_into_mongo, process_data". In general, you can usually delete the asterisk in your IDE and then ask it for help with autocomplete of specific symbols to import. Or go down to the call site, marked Red for Undefined, and right-click will help you adjust the `import`.

src/main.py
smaysenhalder: I think this is the code that was lost which I added back:

q = wiki_to_netflix.format_sparql_query('The Room', 2003)
print(q)

audiodude: Sorry I meant this comment:

```
# Test is true if no argument is passed or if the first argument is not '--prod'.
```

But now I see that the comment is still in the `wiki_to_netflix.py` script, it just didn't get copied over when the check for `--prod` got copied. No big deal.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -4,12 +4,15 @@
 import time
 from tqdm import tqdm
 import sys
+from db.insert_data import insert_into_mongo
```
audiodude, line 7: Test is failing:

```
ModuleNotFoundError: No module named 'db'
```

Should be `mediabridge.db.insert_data`. But actually, are you using this in this file?

audiodude (approved):
Great work. Just a couple tiny things that could or could not be done in this PR. Ready to merge as-is.

mediabridge/data_processing/wiki_to_netflix.py
smaysenhalder: Just updated, the import is to be able to do: data = process_data(test = test) but I don't think insert_data exists anymore. I don't remember who did the process_data(test = test) part so I'm not 100% sure what it does.

audiodude: Okay, so `process_data` is a function in this file. But I don't understand why you're worried about that function, because this import is for `insert_into_mongo`. We can quickly search this file (`wiki_to_netflix.py`) and see that it's not being used, so it can safely be removed.

jhanley634:
This branch is definitely moribund at this point. It has drifted away from `main`, and has merge conflicts in db/ and data_processing/ that I don't know how to properly resolve. Additionally the new .vscode/{extensions,settings}.json files would perhaps be better added to .gitignore. Consider abandoning and deleting this branch.

audiodude:
I tried rebasing this branch in order to merge it, but it's very out of date at this point.

The work to establish a `constants.py` has been sort of superseded by the work in #54.

We still need the Mongo code, but it should probably just be done in its own new PR at this point.

PR #22, opened 2024-10-17
Fix generator error, update data and output directories

cocomittens:
Fixes issue w trying to get the length of a generator which breaks everything :(
Also updates the input/output directories so that everything goes where it should

### DIFF
--- a/src/data_processing/wiki_to_netflix.py
+++ b/src/data_processing/wiki_to_netflix.py
@@ -114,17 +114,17 @@ def process_data(test=False):
     missing_count = 0
     processed_data = []
 
-    netflix_file = read_netflix_txt(os.path.join(data_dir, 'movie_titles.txt'), test)
+    netflix_data = read_netflix_txt(os.path.join(data_dir, 'movie_titles.txt'), test)
 
     netflix_csv = os.path.join(out_dir, 'movie_titles.csv')
 
-    wiki_movie_ids_list, wiki_genres_list, wiki_directors_list = wiki_query(netflix_file, user_agent)
+    wiki_movie_ids_list, wiki_genres_list, wiki_directors_list = wiki_query(netflix_data, user_agent)
 
     num_rows = len(wiki_movie_ids_list)
 
-    netflix_file = read_netflix_txt(os.path.join(data_dir, 'movie_titles.txt'), test)
+    netflix_data = read_netflix_txt(os.path.join(data_dir, 'movie_titles.txt'), test)
 
-    for index, row in enumerate(netflix_file):
+    for index, row in enumerate(netflix_data):
         netflix_id, year, title = row
         if wiki_movie_ids_list[index] is None:
             missing_count += 1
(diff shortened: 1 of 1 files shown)


src/data_processing/wiki_to_netflix.py
```diff
@@ -115,14 +114,17 @@ def process_data(test=False):
     missing_count = 0
     processed_data = []
 
-    netflix_data = read_netflix_txt(os.path.join(data_dir, 'movie_titles.txt'), test)
-    num_rows = len(netflix_data)
+    netflix_file = read_netflix_txt(os.path.join(data_dir, 'movie_titles.txt'), test)
```
audiodude, line 117: It looks like you accidentally renamed this to`netflix_file`. It should be `netflix_data` right?

src/data_processing/wiki_to_netflix.py
```diff
@@ -115,14 +114,17 @@ def process_data(test=False):
...
 
     netflix_csv = os.path.join(out_dir, 'movie_titles.csv')
 
-    wiki_movie_ids_list, wiki_genres_list, wiki_directors_list = wiki_query(netflix_data, user_agent)
+    wiki_movie_ids_list, wiki_genres_list, wiki_directors_list = wiki_query(netflix_file, user_agent)
+
+    num_rows = len(wiki_movie_ids_list)
+
+    netflix_file = read_netflix_txt(os.path.join(data_dir, 'movie_titles.txt'), test)
```
audiodude, line 125: Looks like this is duplicated from above. I think you want to remove this line?

audiodude (changes requested):
Good idea to do PRs with small changes, I like it!

audiodude:
Also regarding your comments on how this code blows up because of the generator. To me, that says that no one ran the code before it was committed (not naming names). I think #15 becomes more important already.

cocomittens:
> Also regarding your comments on how this code blows up because of the generator. To me, that says that no one ran the code before it was committed (not naming names). I think #15 becomes more important already.

Yea I have committed QA sins, I shall atone for them/ fix the above mentioned comments
I happen to also be working on [a branch](https://github.com/noisebridge/MediaBridge/tree/add-tests) that seems like it at least somewhat works. Will improve/ add more tests and probably integrate them into the ci/cd pipeline with some kinda github action (in a separate pr) to prevent future me from accidentally contaminating the codebase with impure untested code.

audiodude:
> > Also regarding your comments on how this code blows up because of the generator. To me, that says that no one ran the code before it was committed (not naming names). I think #15 becomes more important already.
> 
> Yea I have committed QA sins, I shall atone for them/ fix the above mentioned comments I happen to also be working on [a branch](https://github.com/noisebridge/MediaBridge/pull/23) that seems like it at least somewhat works. Will improve/ add more tests and probably integrate them into the ci/cd pipeline with some kinda github action (in a separate pr) to prevent future me from accidentally contaminating the codebase with impure untested code.

So much drama 😅😂

src/data_processing/wiki_to_netflix.py
cocomittens: Yes I believe you are correct 😅 Fixed

src/data_processing/wiki_to_netflix.py
cocomittens: Yup it has been removed

audiodude (approved):
LGTM

PR #23, opened 2024-10-17
Add first basic test

cocomittens:


tests/constants.py
```diff
@@ -0,0 +1,45 @@
+SPARQL_QUERY ='''
```
cocomittens, line 1: IK copy can change so this probably isnt the best approach to just have it here like this, but just testing the tests tbh

audiodude: Yeah we sometimes call data like this "golden data" because it's a specific snapshot of the output of some code. As you probably guessed, if we update the SPARQL query we'd have to update this golden as well.

You don't need to put it in its own file, it's fine to just have it at the top of the test file where it is used.

cocomittens: Yeah for sure, will update this next

I agree that it could be put there, but I would argue that it would be ugly so perhaps it shouldn't be put there. When I look at my test file I wanna see tests, not this 40+ lines SPARQL query

audiodude: I understand what you mean. The tradeoff is "cluttering a test where the data is used" versus "cluttering the file system with modules", since you're only going to use this variable in the specific test file. When I look at this `constants.py` file `(which` should probably be called `goldens.py` or `test_data.py`), it's not immediately clear what it is or where it's used. Keeping the test data close to the tests is generally a good practice.

Also, nit, I would name it `EXPECTED_SPARQL_QUERY`, since it is what you expect your function under test to output, not something that is used as a SPARQL query on its own.

tests/__pycache__/__init__.cpython-312.pyc
audiodude, line 1: You accidentally committed a `__pycache__` folder, which is used by Python to cache parsed source files. It should never be committed. Please delete and add `__pycache__` to `.gitignore`.

tests/constants.py
audiodude, line 1: My personal preference is to put tests directly next to source files. So if you have `src/data/processing/wiki_to_netflix.py` you would also have `src/data/processing/wiki_to_netflix_test.py`. This makes it easier to tell which files have or are missing tests, as well as easier to navigate the source tree when "flipping back and forth" between the source file and the test file.

 You can configure pytest to look for files with a `_test.py` extension.

Let me know what you think.

audiodude (changes requested):

tests/__pycache__/__init__.cpython-312.pyc
cocomittens: Fixed

tests/constants.py
cocomittens: Fixed

tests/constants.py
cocomittens: Fixed

cocomittens:
Ran the test w `pytest` & it works

skyfenton (commented):
Nit: I'd prefer test_\*.py as the naming convention so that if we have long files names that get abbreviated, we can tell which one is the test file without scrolling to the end, but test_data_*.py seems even worse. Thoughts? 

Maybe we use like _test_*.py for the file with the misc long test data or some other symbol (!test_) ?

cocomittens:
> Nit: I'd prefer test__.py as the naming convention so that if we have long files names that get abbreviated, we can tell which one is the test file without scrolling to the end, but test_data__.py seems even worse. Thoughts?
> 
> Maybe we use like _test_*.py for the file with the misc long test data or some other symbol (!test_) ?

I see your point, but I feel like the file name being first more easily shows which file it is testing. Plus I feel like the filenames are automatically sorted alphabetically, and it seems like not having additional letters automatically gets ranked first. So the even if we can't see the whole filename, the file itself we can assume is first, then filename + '_test', then filename + _test_data'. Also, if test was in front of all the tests I feel like they would be grouped together as opposed to automatically placed near the files they are testing, in which case might as well just have a test folder.

skyfenton (approved):
LGTM

mediabridge/data_processing/__pycache__/wiki_to_netflix.cpython-312.pyc
audiodude, line 1: Please remove remaining __pycache__ file.

audiodude (changes requested):

mediabridge/data_processing/__pycache__/wiki_to_netflix.cpython-312.pyc
skyfenton: I don't actually see it in the branch?

skyfenton (changes requested):
Forgot to request per #15: document in the README how to run the tests ('pipenv run pytest', though I would say working in the env with pipenv shell would be easier?).

mediabridge/data_processing/__pycache__/wiki_to_netflix.cpython-312.pyc
audiodude: This comment is on the file, in the changeset, so it's definitely there.

audiodude:
> Forgot to request per #15: document in the README how to run the tests ('pipenv run pytest', though I would say working in the env with pipenv shell would be easier?).

Though using the pipenv shell takes more explaining

mediabridge/data_processing/__pycache__/wiki_to_netflix.cpython-312.pyc
cocomittens: I removed the file and pushed the changes. I think the change may be deletion of the file that seems to currently be on [main](https://github.com/noisebridge/MediaBridge/tree/main/mediabridge/data_processing/__pycache__). Also I think the minus sign next to it might mean deleted.

cocomittens:
> Forgot to request per #15: document in the README how to run the tests ('pipenv run pytest', though I would say working in the env with pipenv shell would be easier?).

Done

mediabridge/data_processing/__pycache__/wiki_to_netflix.cpython-312.pyc
audiodude: You're right, this is being deleted from the main branch. Thanks!

audiodude (approved):

PR #25, opened 2024-10-18
New directory structure

audiodude:
After this, we will run main as `pipenv run python -m mediabridge.main`. All imports will start with `mediabridge`, eg `mediabridge.data_processing.database.foo.constants`


skyfenton (approved):
LGTM

PR #33, opened 2024-10-26
Add ruff for code linting/formatting

### LINKED ISSUE #18: Figure out project standards for code linting, formatting and import order
Probably the best choice is [black](https://black.readthedocs.io/en/stable/index.html), which is used by lots of the official Python projects. There's also [ruff](https://docs.astral.sh/ruff/) which is supposedly faster, but I doubt it will matter for the handful of Python files we'll have. Not sure if either of these provide import formatting, I know about [isort](https://pycqa.github.io/isort/) for that.

Also as far as enforcing these, we can just trust everyone to install the formatters in their VSCode or other editors, or we can create git "hooks" that enforce it every time you commit (but that requires setting up a library like [pre-commit](https://pre-commit.com/) which can be bypassed), or we can set up a GitHub action that tries to do the formatting and fails if there are any changes. The GitHub action would be configured as a "check" so that PRs that violate it (aren't formatted) wouldn't be able to be checked in.

Discuss your ideas and opinions here.

audiodude:
This PR adds [ruff](https://docs.astral.sh/ruff/) as the project's code linter and formatter for Python files. Currently the configuration, in the `ruff.toml` file, is the default except with import sorting added.

As part of this PR, all code has been formatted and re-saved.

We also add a `.vscode` folder, so that we can "recommend" the ruff formatter extension, and configure the project to use ruff for Python files, format on save, and sort imports on save.

Finally, we add a CI check that the code passes formatting before it can be checked in.

Fixes #18.


skyfenton (approved):
Looks great and def needed!

PR #35, opened 2024-11-01
Added Docstrings to wiki_to_netflix.py

JamesKohlsRepo:
Main function still needs complete docstring. Requests in pipfle was changed to wildcard, we don't need to commit that change if we need that specific version


audiodude (approved):
This is awesome! Wow, very comprehensive, great job!

I would revert the Pipfile changes though, I already have #32 to loosen the requirements.

audiodude:
Also you need to merge main so you get the workflow for ruff, which is required to merge. Once you have the workflow (in `.github/workflows/ci`), the ruff task will run and we can merge.

PR #37, opened 2024-11-01
Replace comments with Docstrings

JamesKohlsRepo:
Replaced comments with Docstrings, replying previous changes due to issues with main brach

### DIFF
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -79,8 +79,7 @@ def wiki_feature_info(data, key):
     Returns:
         None: If the key is not present or no results are available
         List: If the key is 'genreLabel', returns a list of unique genre labels.
-        String: If they Key is present, return the movie ID of the firt binding, in other words the first row in query result
-
+        String: If they Key is present, return the movie ID of the first binding, in other words the first row in query result
     """
     if (
         len(data["results"]["bindings"]) < 1
@@ -107,7 +106,7 @@ def format_sparql_query(title, year):
     year (int): release year of the movie
 
     Returns:
-    SPARKLE Query (str): formatted string with movie title and year
+    SPARQL Query (str): formatted string with movie title and year
     """
     QUERY = """
         SELECT * WHERE {
@@ -219,10 +218,7 @@ def wiki_query(data_csv, user_agent):
 
 def process_data(test=False):
     """
-    Calls all functions ...
-
-    Parameters:
-    test (bool): ...
+    If true, only process a subset of the data, for testing
     """
     missing_count = 0
     processed_data = []
@@ -262,7 +258,5 @@ def process_data(test=False):
 
 if __name__ == "__main__":
     # Test is true if no argument is passed or if the first argument is not '--prod'.
-    test = len(sys.argv) < 2 or sys.argv[1] != "--prod"
-    process_data(test=test)
     test = len(sys.argv) < 2 or sys.argv[1] != "--prod"
     process_data(test=test)
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -79,7 +79,7 @@ def wiki_feature_info(data, key):
     Returns:
         None: If the key is not present or no results are available
         List: If the key is 'genreLabel', returns a list of unique genre labels.
-        String: If they Key is present, return the movie ID of the first binding, in other words the first row in query result
+        String: If the Key is present, return the movie ID of the first binding, in other words the first row in query result
     """
     if (
         len(data["results"]["bindings"]) < 1
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -25,7 +25,7 @@ def read_netflix_txt(txt_file, test):
     test (Bool): When true, runs the functon in test mode
     """
     num_rows = None
-    if test == True:
+    if test:
         num_rows = 100
 
     with open(txt_file, "r", encoding="ISO-8859-1") as netflix_data:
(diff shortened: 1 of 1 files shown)


mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -190,4 +264,5 @@ def process_data(test=False):
     # Test is true if no argument is passed or if the first argument is not '--prod'.
     test = len(sys.argv) < 2 or sys.argv[1] != "--prod"
     process_data(test=test)
+    test = len(sys.argv) < 2 or sys.argv[1] != "--prod"
```
audiodude, line 267: nit: Looks like a duplicate line from a merge? Please remove.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -148,8 +217,13 @@ def wiki_query(data_csv, user_agent):
     return wiki_movie_ids, wiki_genres, wiki_directors
 
 
-# Calling all functions
 def process_data(test=False):
+    """
+    Calls all functions ...
+
+    Parameters:
+    test (bool): ...
```
audiodude, line 225: Test is "If true, only process a subset of the data, for testing"

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -53,8 +98,17 @@ def wiki_feature_info(data, key):
...
 def format_sparql_query(title, year):
+    """
+    Formats SPARQL query for Wiki data
+
+    Parameters:
+    title (str): name of content to query
+    year (int): release year of the movie
+
+    Returns:
+    SPARKLE Query (str): formatted string with movie title and year
```
audiodude, line 110: nit: SPARQL

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -29,14 +35,53 @@ def read_netflix_txt(txt_file, test):
...
+                    },
+                ]
+            }
+        }
+    key (str): The key for the information to extract (e.g. 'item', 'genreLabel', 'directorLabel').
+
+    Returns:
+        None: If the key is not present or no results are available
+        List: If the key is 'genreLabel', returns a list of unique genre labels.
+        String: If they Key is present, return the movie ID of the firt binding, in other words the first row in query result
```
audiodude, line 82: nit: s/firt/first/

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -16,10 +16,16 @@ class WikidataServiceTimeoutException(Exception):
...
+    """
+    Reads and processes a Netflix text file.
+
+    Parameters:
+    txt_file (str): Path to the Netflix text file
+    test (Bool): When true, runs the functon in test mode
+    """
     num_rows = None
-    if test:
+    if test == True:
```
audiodude, line 28: nit: bad merge? Please restore to `if test:`

audiodude (changes requested):
Still looks great in it's new reincarnation, but I think there are some issues with the merge. Please double check the comments I've left, thanks!

audiodude:
The ruff error can be fixed by applying the suggestions in my comments. Thanks!

mediabridge/data_processing/wiki_to_netflix.py
JamesKohlsRepo: Will do thanks for catching that

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -79,8 +79,7 @@ def wiki_feature_info(data, key):
     Returns:
         None: If the key is not present or no results are available
         List: If the key is 'genreLabel', returns a list of unique genre labels.
-        String: If they Key is present, return the movie ID of the firt binding, in other words the first row in query result
-
+        String: If they Key is present, return the movie ID of the first binding, in other words the first row in query result
```
audiodude, line 82: nit: s/if they Key/if the Key/

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -148,8 +216,10 @@ def wiki_query(data_csv, user_agent):
     return wiki_movie_ids, wiki_genres, wiki_directors
 
 
-# Calling all functions
 def process_data(test=False):
+    """
+    If true, only process a subset of the data, for testing
```
audiodude, line 221: how about:

```
Reads movie titles and release years from the Netflix data set (which should be downloaded and placed in the repo), then tries to match them with data from Wikidata. For any matches, a CSV file is written.
```

jhanley634 (approved):
Much of this looks good. Sorry about the merge nonsense, you'll need to tidy that up. Going forward, you (roughly) likely want to have at most two feature branches outstanding -- an old one that is out for review and a new one to work on while awaiting final review. The repo has a bunch of branches currently, which makes it a little harder to schedule work so no one steps on another's toes. Also, putting an issue number in the branch name (like "16-add-logging") can help to chronologically order them, and identify branches which have gone stale and maybe should be culled.

----

The wiki_feature_info() docstring is harder to read than it ought to be. The example dict is great but it takes up a lot of vertical space. Better to break out `def get_example_response()` having a single statement, which will `return` a constant, a dict. Several good things would come of that:

1. Humans can easily see there's a pair of input params, in a tabular list, readily scanned.
2. IDEs better support auto-indent when the material is code rather than plain text.
3. Unit tests can operate on the example record, without need of a network connection.

----

At the call site we see three parallel datastructures:
```
        wiki_movie_ids.append(wiki_feature_info(data, "item"))
        wiki_genres.append(wiki_feature_info(data, "genreLabel"))
        wiki_directors.append(wiki_feature_info(data, "directorLabel"))
```
It isn't always apparent to a maintenance engineer *whether*, and *how many* such lists are connected to one another. It is out-of-scope for this one, but in a subsequent PR you might consider defining a three-field [@dataclass](https://docs.python.org/3/library/dataclasses.html), to consolidate all this into just a single list. If for some reason you're not keen on that, there is always [NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple), or even the lowly dict.

audiodude (approved):

audiodude:
I agree with @jhanley634 that parallel lists is an anti-pattern, see https://codeblog.jonskeet.uk/2014/06/03/anti-pattern-parallel-collections/

PR #38, opened 2024-11-02
Anti-pattern correction and Docstring Adjustments 

JamesKohlsRepo:
Trying to improve clarity of Docstrings recommended by previous pr. Working on removing [anti-patterns ](https://codeblog.jonskeet.uk/2014/06/03/anti-pattern-parallel-collections/). Peer programming would be nice to clean up the rest of wiki_to_netflix.py

### DIFF
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -13,7 +13,7 @@ class WikidataServiceTimeoutException(Exception):
 
 
 @dataclass
-class movieData:
+class MovieData:
     movie_id: str
     genre: str
     director: str
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -2,7 +2,7 @@ import csv
 import os
 import sys
 import time
-from dataclasses import dataclass, field
+from dataclasses import dataclass
 from typing import List, Optional
 
 import requests
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -2,6 +2,7 @@ import csv
 import os
 import sys
 import time
+from dataclasses import dataclass
 
 import requests
 from tqdm import tqdm
@@ -9,6 +10,13 @@ from tqdm import tqdm
 
 class WikidataServiceTimeoutException(Exception):
     pass
+
+
+@dataclass
+class movieData:
+    movie_id: str
+    genre: str
+    director: str
 
 
 data_dir = os.path.join(os.path.dirname(__file__), "../../data")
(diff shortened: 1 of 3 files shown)


mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -11,6 +12,13 @@ class WikidataServiceTimeoutException(Exception):
     pass
 
 
+@dataclass
+class movieData:
```
audiodude, line 16: Classes, even dataclasses, should start with a capital letter.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -209,11 +185,15 @@ def wiki_query(data_csv, user_agent):
         response.raise_for_status()
         data = response.json()
 
-        wiki_movie_ids.append(wiki_feature_info(data, "item"))
-        wiki_genres.append(wiki_feature_info(data, "genreLabel"))
-        wiki_directors.append(wiki_feature_info(data, "directorLabel"))
+        wiki_data_list.append(
```
audiodude, line 188: Very important, you also need to update the `process_data` function, where this is called, and have it handle the data in the new format.

audiodude (changes requested):
Good idea, but make sure you update the rest of the program with your changes!

mediabridge/data_processing/wiki_to_netflix.py
jhanley634: @JamesKohlsRepo , your IDE is a power tool. Take advantage of it! Practice makes perfect.

Renaming is a pretty common refactor, and not just for CamelVersus_snake_case minutiae. Commonly we will "guess" at what something means, write code, then realize it means something a little different so we should rename to reflect that. Begin by putting the cursor on the identifier we wish to change.

In Pycharm, right-click to get a context menu, choose Refactor => Rename, adjust to the new spelling of the identifier, hit RETURN.

In VSCode, right-click, choose Rename, type the name, hit RETURN.

The IDE will then chase through the codebase to adjust the definition plus all calling sites, including `import` statements. Use `$ git diff` to verify that things went as planned, then commit the edit.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -209,50 +188,67 @@ def wiki_query(data_csv, user_agent):
...
-    Reads movie titles and release years from the Netflix data set (which should be downloaded and placed in the repo),
-    then tries to match them with data from Wikidata. For any matches, a CSV file is written.
+    Processes Netflix movie data by enriching it with information from Wikidata and writes the results to a CSV file.
+    Netflix data was conveted from a generator to a list to avoid exaustion. was running into an issue where nothing would print to CSV file
     """
     missing_count = 0
     processed_data = []
 
-    netflix_data = read_netflix_txt(os.path.join(data_dir, "movie_titles.txt"), test)
+    netflix_data = list(
```
audiodude, line 210: This is fine for now, but instead of "rendering" the generator into a list (so that it can be iterated through multiple times, which is what it seems we are doing below), we should instead write the code so that it can operate on the data iteratively and not have to reference the rows multiple times. No need to do that here, something to keep in mind for a future PR.

audiodude (approved):

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Also nice job figuring out why the CSV wasn't being written (though in the big picture, we don't need the CSV).

PR #40, opened 2024-11-02
Update README with running instructions

### LINKED ISSUE #39: Main.py ModuleNotFoundError: No module named 'mediabridge'
I don't think the import is working with the main.py file on the main branch, but idk if other machines are different.

```
Traceback (most recent call last):
  File "/workspaces/MediaBridge/mediabridge/main.py", line 1, in <module>
ModuleNotFoundError: No module named 'mediabridge'
```

Obviously `from data_processing import ___` works, but I know we prefer the `mediabridge.*` format and I don't know if we're just missing something (other than moving the main.py file out of the directory).

audiodude:
Fixes #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,16 @@ To install the project dependencies, first install pipenv globally with `pip ins
 
 To run code in the project, prefix your command with `pipenv run`, a la `pipenv run python -m mediabridge.main`.
 
+## Running main
+
+The "main.py" script is part of the `mediabridge` module. Additionally, running it requires `pipenv run` as mentioned above. So the full command to run the main script (or any other script in the `mediabridge` module) is:
+
+```
+pipenv run python -m mediabridge.main
+```
+
+This should be run from the root of the project directory.
+
 ## Testing
 
 To run unit tests,
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ This should be run from the root of the project directory.
 
 ### Running from VSCode
 
-If you'd like to run the main script (or any other script) from within VSCode, make sure you've let VSCode know about your pipenv environment. To do that:
+To fix import errors and other Intellisense features, make sure you've let VSCode know about your pipenv environment. To do that:
 
-1. Open the VSCode command pallet (Control/Command+SHIFT+P)
-1. Start typing "Python select interpreter" and run the command that pops up
-1. Cboose the option that starts with `MediaBridge`.
+1. Open the VSCode command palette (Control/Command+SHIFT+P)
+2. Search for and select the "Python: Select Interpreter" command
+3. Choose the option that starts with `MediaBridge`
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


skyfenton (commented):
Pipenv run seems like the best way to work, but maybe we should probably also remind people to set their interpreters to their pipenv environment once installed to sync their IDE?

skyfenton:
Also super inconsequential but do people tend to have a shortcut to run python commands like an `npm run dev` type situation? Seems like it could be a little more convenient somehow but it could also be unnecessary.

audiodude:
> Pipenv run seems like the best way to work, but maybe we should probably also remind people to set their interpreters to their pipenv environment once installed to sync their IDE?

I don't run Python code from the IDE, so I'm not entirely sure what the instructions would be, but I think I've seen the `Select Python interpreter...` command before, is that what you're referring to. I can definitely add that.

> Also super inconsequential but do people tend to have a shortcut to run python commands like an `npm run dev` type situation? Seems like it could be a little more convenient somehow but it could also be unnecessary.

See above. Maybe ask in Discord?

skyfenton:
> I don't run Python code from the IDE, so I'm not entirely sure what the instructions would be, but I think I've seen the Select Python interpreter... command before, is that what you're referring to? I can definitely add that.

Yeah exactly, I don't know if it's particularly different across other IDEs, so any generic comment about switching the interpreter should do.

> See above. Maybe ask in Discord?

Guess we'll see if discord has any ideas.

skyfenton:
I just committed some fixes to the file to be more specific/fix typos.

jhanley634:
> Also super inconsequential but do people tend to have a shortcut to run python commands like an `npm run dev` 

Simple instructions are a big win, they _are_ consequential. I offer [this example](https://github.com/jhanley634/dojo-blackboard/blob/7d45d03f/Makefile#L12), using an approach that worked well with more than a dozen teams of collaborators. It lets us write concise ReadMe instructions, like

- make lint
- make test
- make lint test

Short commands are easy to verbally pronounce during a pair debugging session. If the underlying details change, like we move from flake8 to ruff, they are updated in one place, the Makefile, without changes to the docs or to a team member's muscle memory.

My $(ACTIVATE) idiom does not mention `pipenv`, but it could be trivially adjusted.

audiodude:
image

jhanley634:
> "no make"

Meh, whatever, it's an example that has been tested and found to work. Could be a Bash script or similar. Perhaps with `source common.sh` to deal with DRY, perhaps with `set -x` so there is "no magic". Or another language, such as python.

The point I was hoping to make is that "documentation costs" are real, as they affect speed of onboarding new team members, and uniformity of how colleagues actually interact with a codebase. When an engineer `cd`s into a directory and types `ls`, sometimes it is very self-explanatory what commands they should type there. I have obtained good results with colleagues of diverse skill levels by siphoning out as much as feasible from the ReadMe prose and burying most of it down in an executable script.

JamesKohlsRepo (approved):
lgtm!

PR #46, opened 2024-11-08
Add Pipfile scripts for dev and test

### LINKED ISSUE #41: Add pipenv script shortcuts
We were discussing in #40 the possibility of adding shortcuts to running tests, the main file, etc. Turns out Pipenv already has npm style scripts, so I think we should add "pipenv run init/prod/dev/tests"  shortcuts.

Of course, we could do this in make and maybe that could help when we start building the full stack, but I don't think it totally matters one way or the other right now, and just to reduce build steps, I think pipenv scripts work fine.

Here's a link to the [documentation](https://pipenv.pypa.io/en/latest/scripts.html).

skyfenton:
Fixes #41. Also moves ruff package into dev packages b/c not required to run script.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -15,3 +15,6 @@ pytest-cov = "~=5.0"
 
 [requires]
 python_version = "3.12"
+
+[scripts]
+dev = "pipenv run python -m mediabridge.main"
--- a/Pipfile
+++ b/Pipfile
@@ -6,12 +6,12 @@ name = "pypi"
 [packages]
 python-dotenv = "~=1.0"
 requests = "~=2.26"
-ruff = "~=0.7"
 tqdm = "~=4.66"
 
 [dev-packages]
 pytest = "~=8.3"
 pytest-cov = "~=5.0"
+ruff = "~=0.7"
 
 [requires]
 python_version = "3.12"
--- a/Pipfile
+++ b/Pipfile
@@ -18,4 +18,4 @@ python_version = "3.12"
 
 [scripts]
 dev = "pipenv run python -m mediabridge.main"
-tests = "pipenv run pytest"
+test = "pipenv run pytest"
--- a/Pipfile
+++ b/Pipfile
@@ -18,3 +18,4 @@ python_version = "3.12"
 
 [scripts]
 dev = "pipenv run python -m mediabridge.main"
+tests = "pipenv run pytest"
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # What is MediaBridge?
 
-MediaBridge is a project being developed at the [Noisebridge](https://github.com/noisebridge) hackerspace in San Francisco, CA, USA. See also the [Noisebridge hompage](https://www.noisebridge.net/wiki/Noisebridge) and the [wiki entry for this project](https://www.noisebridge.net/wiki/Python_Project_Meetup).
+MediaBridge is a project being developed at the [Noisebridge](https://github.com/noisebridge) hackerspace in San Francisco, CA, USA. See also the [Noisebridge homepage](https://www.noisebridge.net/wiki/Noisebridge) and the [wiki entry for this project](https://www.noisebridge.net/wiki/Python_Project_Meetup).
 
 MediaBridge is in a _very_ early stage of the development. It's intended functionality is to provide recommendations that _bridge_ media types. So for example, you might say you're interested in the film _Saw_ and MediaBrige might recommend the video game _Silent Hill_ or a Stephen King book. For now, we are working on simply returning recommendations for movies, based on the [Netflix Prize dataset](https://www.kaggle.com/datasets/netflix-inc/netflix-prize-data).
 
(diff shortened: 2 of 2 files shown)


jhanley634 (approved):
I really like the "encounter a ModuleNotFoundError" advice, it's trying to be helpful to some poor soul got an error and remembered to consult the docs.

But another approach (a "belt + suspenders" approach) to being helpful is to make the displayed error more diagnostic. You might consider turning "pipenv run pytest" into something like "pipenv run (cd ../MediaBridge && pytest)". If you're already at top of repo, it is idempotent, it's a no-op. If you are elsewhere, the `cd` will complain in a way which reminds the user to chdir to the right spot before trying the command.

skyfenton:
> But another approach (a "belt + suspenders" approach) to being helpful is to make the displayed error more diagnostic. You might consider turning "pipenv run pytest" into something like "pipenv run (cd ../MediaBridge && pytest)". If you're already at top of repo, it is idempotent, it's a no-op. If you are elsewhere, the `cd` will complain in a way which reminds the user to chdir to the right spot before trying the command.

That does sound useful, maybe there's something more resilient we could use in the weird case a user renames their project folder? Also I assume there's a way around it, but wouldn't that conflict with any command line arguments we'll eventually want to pass to main.py? (I'm currently working on using typer to add a verbose flag for logging)

skyfenton:
> But another approach (a "belt + suspenders" approach) to being helpful is to make the displayed error more diagnostic. You might consider turning "pipenv run pytest" into something like "pipenv run (cd ../MediaBridge && pytest)". If you're already at top of repo, it is idempotent, it's a no-op. If you are elsewhere, the `cd` will complain in a way which reminds the user to chdir to the right spot before trying the command.

That does sound useful, maybe there's something more resilient we could use in the weird case a user renames their project folder? Also I assume there's a way around it, but wouldn't that conflict with any command line arguments we'll eventually want to pass to main.py? (I'm currently working on using typer to add a verbose flag for logging)

jhanley634:
> more resilient we could use in the weird case a user renames their project folder?

Certainly a user _could_ rename the folder. But in that case I would view it as, "you are on your own, venturing into unsupported territory." The only reason _I_ can imagine for wanting to have multiple media bridge repos is so you can clone a clean one and compare. In which case I would advise to (cd /tmp && git clone ...), or `mkdir -p /tmp/2nd_test` and clone under that.

----

# args

This project won't be using "make". Not sure why, but it won't. My experience in a bunch of projects is that `$ make test VERBOSE=-v` is a convenient way to pass an arbitrary value down the call stack. And then a recipe line like ` pytest $(VERBOSE)` will run tests with or without verbosity, as the caller requested.

Clearly bash scripts can be pressed into service, accepting args directly, or via env vars.

I have never used pipenv, and haven't run any tests, so I am not yet aware of the argv conventions used by that community. But I _have_ to believe it's straightforward to transmit arbitrary arguments down to a command which consumes them. Ask google or Claude or whatever, test, commit, and see how it goes.

Pipfile
```diff
@@ -6,12 +6,16 @@ name = "pypi"
...
 [dev-packages]
 pytest = "~=8.3"
 pytest-cov = "~=5.0"
+ruff = "~=0.7"
 
 [requires]
 python_version = "3.12"
+
+[scripts]
+dev = "pipenv run python -m mediabridge.main"
```
audiodude, line 20: I think that `main` would be a better name. Just looking through the README changes, it's a weird sentence: "pipenv run dev runs the main script".

audiodude (approved):
This is cool, I didn't realize how simple it would be. Thanks for updating the README as well!

Pipfile
skyfenton: I figured eventually we'd add a "prod" script, so I didn't want to name the dev script "main" since I feel the difference between "main" and "prod" is more ambiguous than "dev" and "prod".

skyfenton:
> This project won't be using "make". Not sure why, but it won't. My experience in a bunch of projects is that `$ make test VERBOSE=-v` is a convenient way to pass an arbitrary value down the call stack. And then a recipe line like ` pytest $(VERBOSE)` will run tests with or without verbosity, as the caller requested.

Trevor and I were discussing whether the project should use make vs pipenv scripts/aliases and we came to the conclusion that pipenv scripts would be a little more frictionless, since everyone has it already when they setup their pipenv environment and we don't have any chain of dependencies we need to manage which make would be useful for. This way we can also increase accessibility to the project by avoiding the roughness around using make on windows if anyone ends up joining the project on a windows machine. I'd def be down to reconsider this if there are cases where make makes aliasing commands even easier.

> I have never used pipenv, and haven't run any tests, so I am not yet aware of the argv conventions used by that community. But I _have_ to believe it's straightforward to transmit arbitrary arguments down to a command which consumes them. Ask google or Claude or whatever, test, commit, and see how it goes.
> 
> I habitually fork off a subshell with (cd /some/where && run_command.sh), but perhaps it's not necessary here. Perhaps removing the `(` `)` parens makes it easier to tack on user-supplied args.

Pipenv sends down args to whatever command you alias, so the current config works with args. I just wasn't sure if I needed to do anything special when using `&&` to avoid passing them to cd or something. I got a sample command somewhat working with the cd trick, but I think documenting the NotFound error alongside the run command might just be the more readable solution.

PR #48, opened 2024-11-14
Add basic logging to wiki to netflixpy

### LINKED ISSUE #16: Add logging to `wiki_to_netflix.py`
We should add some log messages using the Python logging module (https://docs.python.org/3/library/logging.html). Let's try to appropriately use INFO and DEBUG so that we can either get a large amount of messages (DEBUG) or less (INFO).

### LINKED ISSUE #30: Consider using typer
> Clearly the `test` assignment works. But the fact we felt a (helpful!) comment was needed suggests that maybe we're working too hard. Be lazy! Let [typer](https://pypi.org/project/typer) sweat the details about cracking argv. Add a `prod: bool` parameter to the `process_data()` signature. Then you will get `--help` "for free". And no, you certainly don't need to do that in _this_ PR, you might defer that to a subsequent PR.

_Originally posted by @jhanley634 in https://github.com/noisebridge/MediaBridge/issues/17#issuecomment-2438606052_

skyfenton:
Adds basic functionality for logging request success/failure and cleans up some function outputs (the processing_data list now appends a None element if there is no found movie data instead of an empty MovieData object, slightly out of scope, but was already consolidating the edge case). Also adds typer with CLI flags for logging info messages to the console (-v or --verbose) or logging to file in out_dir (--test).

Closes #16 and closes #30.

### DIFF
--- a/mediabridge/main.py
+++ b/mediabridge/main.py
@@ -1,7 +1,6 @@
 import logging
 
 import typer as typer
-from tqdm import trange
 from tqdm.contrib.logging import logging_redirect_tqdm
 
 from mediabridge.data_processing import wiki_to_netflix
@@ -16,12 +15,7 @@ def main(verbose: bool = typer.Option(False, "--verbose", "-v")):
     # Redirect logging to tqdm.write function to avoid colliding with
     # progress bar formatting
     with logging_redirect_tqdm():
-        for i in trange(9):
-            if i == 4:
-                LOG.info("console logging redirected to `tqdm.write()`")
-
-    q = wiki_to_netflix.format_sparql_query("The Room", 2003)
-    print(q)
+        wiki_to_netflix.process_data(True)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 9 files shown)


mediabridge/main.py
```diff
@@ -1,4 +1,32 @@
...
+            filemode="a",
+            format="%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s",
+            datefmt="%H:%M:%S",
+        )
+        wiki_to_netflix.process_data(True)
+        return
+
+    if verbose:
+        logging.basicConfig(level=logging.INFO)
+
```
audiodude, line 24: Any consideration for logging in "production"?

audiodude (approved):
Great work, this looks awesome. One basic question, but I think this is ready to merge.

mediabridge/main.py
skyfenton: Sorry for the wait. The --test option (logging all messages to a file) is how I think the production system will log, but I'm sure we can swap the logging method for whatever monitoring system we want to use later.

PR #51, opened 2024-11-22
Implement frontend

cocomittens:
Sets up basic React frontend (w/ Typescript, Tailwind, & shadCN UI components), w/ basic skeleton of potential add movies to generate recommendation page

Screen Shot 2024-11-18 at 8 01 56 PM

## To test
1. `npm i`
2. `npm run dev`


mediabridge-frontend/src/api/movie.js
audiodude, line 1: Maybe skip this for now. Are we really ever going to `deleteMovie`?

audiodude (approved):
This looks great. I think it strikes a good balance between "set some stuff up" and "create a bunch of boilerplate that we're not going to use".

mediabridge-frontend/src/api/movie.js
cocomittens: Good point, if we do it probably won't be from the UI. Deleted the delete function

audiodude: Sorry I meant the whole file. When are we updating movies from the frontend either?

audiodude: I guess the point is, we don't know yet how we will interact with the API data so it might be best just to leave this out, so we don't box ourselves in.

cocomittens: Hmm, good point, you're right, probably wont be creating/ updating either. I did assume that we would need some sort of `getMovie`/ `getMovies`, and just kinda added the rest of them from muscle memory. Its definitely making assumptions about the API data that may be inaccurate, and was there just as a placeholder for the most part. 

I could delete them all, or just keep `getMovie`/`getMovies` if we are going to use those at some point, then update them later. Fine w either.

audiodude: Totally up to you!

audiodude (approved):

JamesKohlsRepo (approved):
Looks awesome! This really ties the project together

PR #52, opened 2024-11-24
Logging fixes

skyfenton:
Fleshed out some details with the cli/logging including creating an output directory if one doesn't exist and naming logs based on when the program is executed instead of just appending to an existing log.

### DIFF
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -266,11 +266,11 @@ def process_data(test=False):
 
     create_netflix_csv(netflix_csv, processed_data)
 
-    print(f"missing:  {missing_count} ({missing_count / num_rows * 100:.2f}%)")
     print(
-        f"found: {num_rows - missing_count} ({(num_rows - missing_count) / num_rows * 100:.2f}%)"
+        f"missing: {missing_count} ({missing_count / num_rows * 100:.2f}%)\n"
+        f"found: {num_rows - missing_count} ({(num_rows - missing_count) / num_rows * 100:.2f}%)\n"
+        f"total: {num_rows}\n",
     )
-    print(f"total: {num_rows}")
 
 
 if __name__ == "__main__":
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -40,7 +40,7 @@ def read_netflix_txt(txt_file, test):
     """
     num_rows = None
     if test:
-        num_rows = 100
+        num_rows = 50
 
     with open(txt_file, "r", encoding="ISO-8859-1") as netflix_data:
         for i, line in enumerate(netflix_data):
--- a/mediabridge/main.py
+++ b/mediabridge/main.py
@@ -6,17 +6,21 @@ from tqdm.contrib.logging import logging_redirect_tqdm
 from mediabridge.data_processing import wiki_to_netflix
 
 
-def main(verbose: bool = typer.Option(False, "--verbose", "-v"), test: bool = False):
-    if test:
+def main(
+    verbose: bool = typer.Option(False, "--verbose", "-v"),
+    log: bool = False,
+    test: bool = False,
+):
+    if log:
         # log all messages to file
         logging.basicConfig(
             level=logging.DEBUG,
-            filename="out/mediabridge.log",
-            filemode="a",
+            filename=f"out/mb_{datetime.now().strftime('%Y%m%d_%H%M%S')}.log",
+            filemode="w",
             format="%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s",
             datefmt="%H:%M:%S",
         )
-        wiki_to_netflix.process_data(True)
+        wiki_to_netflix.process_data(test)
         return
 
     if verbose:
--- a/mediabridge/main.py
+++ b/mediabridge/main.py
@@ -1,11 +1,12 @@
 import logging
-import os
 from datetime import datetime
+from time import sleep
 
 import typer as typer
 from tqdm.contrib.logging import logging_redirect_tqdm
 
 from mediabridge.data_processing import wiki_to_netflix
+from mediabridge.definitions import OUTPUT_DIR
 
 
 def main(
@@ -14,15 +15,19 @@ def main(
     test: bool = False,
 ):
     # create output directory if it doesn't exist
-    if not os.path.exists("./out"):
-        logging.warning(" /out directory does not exist, creating...")
-        os.mkdir("./out")
+    if not OUTPUT_DIR.exists():
+        print(
+            f"[WARNING] Output directory does not exist, creating new directory at {OUTPUT_DIR}"
+        )
+        OUTPUT_DIR.mkdir()
 
     if log:
         # log all messages to file
         logging.basicConfig(
             level=logging.DEBUG,
-            filename=f"out/mb_{datetime.now().strftime('%Y%m%d_%H%M%S')}.log",
+            filename=OUTPUT_DIR.joinpath(
+                f"mb_{datetime.now().strftime('%Y%m%d_%H%M%S')}.log"
+            ),
             filemode="w",
             format="%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s",
             datefmt="%H:%M:%S",
--- a/mediabridge/main.py
+++ b/mediabridge/main.py
@@ -29,12 +29,15 @@ def main(
         return
 
     if verbose:
-        logging.basicConfig(level=logging.INFO)
+        level = logging.INFO
+    else:
+        level = logging.WARNING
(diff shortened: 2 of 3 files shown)


mediabridge/main.py
```diff
@@ -1,31 +1,49 @@
...
 
-def main(verbose: bool = typer.Option(False, "--verbose", "-v"), test: bool = False):
-    if test:
+def main(
+    verbose: bool = typer.Option(False, "--verbose", "-v"),
+    log: bool = False,
+    test: bool = False,
+):
+    # create output directory if it doesn't exist
+    if not os.path.exists("./out"):
```
audiodude, line 17: This will create the `out` directory in whatever directory the script was run from. While we expect that this will usually be the MediaBridge root directory, there are no guarantees.

mediabridge/main.py
```diff
@@ -1,31 +1,49 @@
...
-            filename="out/mediabridge.log",
-            filemode="a",
+            filename=f"out/mb_{datetime.now().strftime('%Y%m%d_%H%M%S')}.log",
+            filemode="w",
             format="%(asctime)s,%(msecs)d %(name)s %(levelname)s %(message)s",
             datefmt="%H:%M:%S",
         )
-        wiki_to_netflix.process_data(True)
+        try:
+            wiki_to_netflix.process_data(test)
```
audiodude, line 31: We shouldn't call `process_data` from two different paths. There should be one place when we call it, which increases readability and future maintainability (what if we want to change something about the call and we forget to do it in both places?).

Instead, you should always catch the exception, and only log it if `log` is `True`.

I guess I missed this in the previous PR, sorry.

mediabridge/main.py
```diff
@@ -1,31 +1,49 @@
...
 
 from mediabridge.data_processing import wiki_to_netflix
 
 
-def main(verbose: bool = typer.Option(False, "--verbose", "-v"), test: bool = False):
-    if test:
+def main(
+    verbose: bool = typer.Option(False, "--verbose", "-v"),
+    log: bool = False,
+    test: bool = False,
```
audiodude, line 14: The old behavior was that test defaulted to `True` and we passed a `--prod` flag to enable the production behavior. I'm fine with it either way.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -266,11 +266,11 @@ def process_data(test=False):
 
     create_netflix_csv(netflix_csv, processed_data)
 
-    print(f"missing:  {missing_count} ({missing_count / num_rows * 100:.2f}%)")
     print(
-        f"found: {num_rows - missing_count} ({(num_rows - missing_count) / num_rows * 100:.2f}%)"
+        f"missing: {missing_count} ({missing_count / num_rows * 100:.2f}%)\n"
```
audiodude, line 270: I don't think merging this into one print statement with `\n` separating them makes it any more readable. There's already a readability challenge in the extensive use of `{}` notation. It is easy to reason that each `print` statement goes on its own line, but harder to notice the `\n` characters in this combination.

audiodude (approved):
Looking good. Just found a few small things but nothing is terribly important.

mediabridge/main.py
skyfenton: I'm aware but not sure how we'd work around this. The 'mediabridge' module imports fail if you aren't using `pipenv run` in the root project directory, so it shouldn't be possible to work around this, but I can't come up with an explicit way to always ensure the main script is run from the root.

skyfenton: We could check if the cwd path ends with MediaBridge but that seems a little silly. And yeah, make does exist.

audiodude: Think about it the other way, how do we ensure that the directory is created in the same place, regardless of where the script is run from.

We can use the `__file__` built-in to get the path to the file we are currently in. If we know that, we know for sure where the `out` directory should be.

In fact, we already do this caclulation [here](https://github.com/noisebridge/MediaBridge/blob/main/mediabridge/data_processing/wiki_to_netflix.py#L29). You should just `mkdir` that variable.

mediabridge/main.py
skyfenton: In wiki_to_netflix.py, the name main and prod flag checks are still there since we talked about still calling the script directly, I think the previous behavior is probably better. Maybe "--full" is a better descriptor than "prod" (since we don't know what our actual production process is going to look like)?

Also, about running script files directly, I think we should find a way so that you can still call verbose/logging flags to other scripts besides main, whether that's just being able to specify a module as an arg to the main script or maybe there's some way to specify in a script's __name__ check that this script has common arguments it should check for with main.

mediabridge/data_processing/wiki_to_netflix.py
skyfenton: I did it just cause it makes the whitespace more consistent so it looks in the code how it does in the output. I can revert this if you still disagree.

mediabridge/main.py
skyfenton: My intention was to keep the calls separated so that we only use the tqdm handler when we're logging to stdout, but I don't think the handler has an impact on perf anyway so I consolidated it all in the latest commit.

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Your choice, just offering my opinion.

mediabridge/main.py
audiodude: I like the idea of a `--full` flag!

The logging setup, a call to `basicConfig` or other more complex configurations, propagates through the whole runtime. What that means is that any logger created with `logging.getLogger()`, from any module, will share the same logging rules. Does that answer your question?

mediabridge/main.py
audiodude: Oh sorry, I'm confused. This is `mediabridge/main.py` not `wiki_to_netflix.py`, where that variable is. In #21, @smaysenhalder is adding a constants file that has the directory paths in it. That would clearly be very useful here. Maybe put a TODO to wait until those constants are ready and skip creating the output directory for now? Or just recalculate the output directory like `wiki_to_netflix.py` does and put a TODO to remove the duplication once it's in the constants file?

mediabridge/main.py
skyfenton: I'll rename it to --full then! :)

Also I meant more along the lines of what we were discussing last meeting where we should keep some other entry points besides main to run individual systems. We could solve that with some arg to main.py like --module=wiki_to_netflix to just run the wiki_to_netflix system while also handling -v or --log, but maybe there's some way to propagate cl args to submodules without duplicating logic, unless you were thinking of a different solution to running individual systems?

audiodude: Yes, we could make the "main" main a kind of wrapper/super module. It could have different commands it could run from other modules, even just calling their `main()` methods.

On the other hand, that kind of conflicts with the `pipenv run <foo>` stuff right? If we want to run module A, presumably we would do `pipenv run A`.

mediabridge/main.py
skyfenton: I don't know how I didn't already know about the __file__ special symbol. I just used it in the latest commit to localize our path definitions in a py file not knowing about the work in #21 to ensure we won't make accidental string manipulation errors.

skyfenton:
using logging_redirect_tqdm actually redirects to stdout when trying to write to a file, working on a fix

mediabridge/main.py
```diff
@@ -1,49 +1,54 @@
...
             datefmt="%H:%M:%S",
         )
+    else:
+        if verbose:
+            level = logging.INFO
+        else:
+            level = logging.WARNING
+        logging.basicConfig(level=level, format="[%(levelname)s] %(message)s")
+
+    # If writing to stdout (log is false), redirect logging to tqdm.write
```
audiodude, line 42: I'm not sure I understand this comment. If we're not logging, we need to redirect tqdm? I thought we only need to redirect it if we *are* logging, so that the tqdm prints don't end up in the log. Or, as per your comment, the logging itself gets redirected to `tqdm.write`? How does that work?

audiodude (commented):
Ending up more confused than I started, please help me out!

mediabridge/main.py
audiodude: I guess I'm looking for a comment along the lines of "tqdm would normally do X. If logging is enabled, use this context manager to make it do Y."

skyfenton: See other comment, but logging still outputs errors and other important things to stdout unless we tell main to log to a file with the --log arg. So we want to redirect log messages that we intend to go to stdout through tqdm so we don't interrupt the formatting, but we do not want to redirect when we're logging to a file (because then tqdm will just steal the messages away from logging's filewriter).

audiodude (approved):
Thanks for the explanations!

skyfenton:
Added those explanations to the code so it's clear for everyone, should be all set to merge afterwards unless there are any last changes.

audiodude (approved):

PR #53, opened 2024-12-06
Easier debugging of SPARQL queries

audiodude:
Add a `--num_rows` argument to more easily specify that value on the command line without having to edit the code. If `--full` is specified, it overrides `--num_rows` and all rows are processed. The `wiki_to_netflix` module still contains a default value for num rows for when it is run directly.

Add a `--missing_out_path` argument, which will collect all the missing movies and put them in their own CSV. If this argument isn't specified, no missing movie CSV is written. This path is relative to the output directory, so if you specify `--missing_out_path=my_missing.csv` you will find your file at `out/my_missing.csv`.

Here is an example of how to use these new arguments:

```
$ pipenv run dev -m missing.old.csv -n 1000
[WARNING] Could not find movie id 1 ('Dinosaur Planet', '2003')
....
missing: 486 (48.60%)
found: 514 (51.40%)
total: 1000

$ pipenv run dev -m missing.new.csv -n 1000
[WARNING] Could not find movie id 1 ('Dinosaur Planet', '2003')
....
missing: 524 (52.40%)
found: 476 (47.60%)
total: 1000

$ wc -l out/missing.old.csv
     486 out/missing.old.csv
$ wc -l out/missing.new.csv
     524 out/missing.new.csv
$ diff out/missing.old.csv out/missing.new.csv
5d4
< 7,null,8 Man,1992,null,null
7d5
< 10,null,Fighter,2001,null,null
16a15
> 28,null,Lilo and Stitch,2002,null,null
27a27
> 44,null,Spitfire Grill,1996,null,null
31a32
> 54,null,We're Not Married,1952,null,null
38d38
...
```

Note that in this example, the QUERY variable in `wiki_to_netflix` was manually edited between run. A future enhancement might be to allow the user to specify a file or python module path that contains the query to use.

### DIFF
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -216,6 +216,9 @@ def process_data(num_rows=None, output_missing_csv_path=None):
     """
     Processes Netflix movie data by enriching it with information from Wikidata and writes the results to a CSV file.
     Netflix data was conveted from a generator to a list to avoid exaustion. was running into an issue where nothing would print to CSV file
+
+    num_rows (int): Number of rows to process. If None, all rows are processed.
+    output_missing_csv_path (str): If provided, movies that could not be matched will be written to a CSV at this path.
     """
     missing_count = 0
     processed_data = []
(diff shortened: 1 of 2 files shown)


skyfenton (commented):
Lgtm. I have a simple change in my branch to use the definitions file for paths in wiki_to_netflix (which also uses pathlib objects instead of os.path functions), but this still works for now.

skyfenton:
Just one question, how do we make use of the missing outputs csv?

audiodude:
> Just one question, how do we make use of the missing outputs csv?

You can see it in my example above. You can use `wc -l <file>` to see how many misses are in each file. You can use `diff` to see which movies are in one file and not the other. You can also presumably import the CSVs into a spreadsheet and compare them that way.

skyfenton (approved):

skyfenton (approved):
Sorry for the delay, looks good! If it's useful to you then it's useful to someone.

PR #54, opened 2024-12-11
Add checks to ensure required files and directories exist

### LINKED ISSUE #24: Put DATA_DIR and OUT_DIR constants in a constants.py file
Instead of resolving the paths from `wiki_to_netflix.py` and possibly having to update them when the paths change, or having to calculate them differently in different files, we should have a `constants.py` file that holds this information across the project.

skyfenton:
Simple changes to refactor the wiki_to_netflix script to use the paths in definitions.py and check that the correct data directories exist.

Also fixes #24 by including paths for the output and data directories in definitions.py. Using these pathlib objects makes it easier and less verbose to perform file operations/checks, but also ensures consistency in pathing across POSIX and win 32.

### DIFF
--- a/mediabridge/data_processing/wiki_to_netflix.py
+++ b/mediabridge/data_processing/wiki_to_netflix.py
@@ -26,8 +26,7 @@ log = logging.getLogger(__name__)
 
 # need Genres, Directors, Title, year?
 
-user_agent = "Noisebridge MovieBot 0.0.1/Audiodude <audiodude@gmail.com>"
-
+USER_AGENT = "Noisebridge MovieBot 0.0.1/Audiodude <audiodude@gmail.com>"
 DEFAULT_TEST_ROWS = 100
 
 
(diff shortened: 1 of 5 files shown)


mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -220,15 +218,30 @@ def process_data(num_rows=None, output_missing_csv_path=None):
     num_rows (int): Number of rows to process. If None, all rows are processed.
     output_missing_csv_path (str): If provided, movies that could not be matched will be written to a CSV at this path.
     """
+
+    if not DATA_DIR.exists():
+        raise FileNotFoundError(
+            f"Data directory does not exist at {DATA_DIR}, please create a new directory containing the netflix prize dataset files\n"
+            "https://archive.org/details/nf_prize_dataset.tar"
```
skyfenton, line 225: We could extract the archive.org link into an environment variable, or just omit entirely depending on how we want to expose the dataset download?

audiodude: Not sure what you mean. Someone was working on automatic download of the Netflix data. I think @JamesKohlsRepo

audiodude (approved):

mediabridge/data_processing/wiki_to_netflix.py
skyfenton: Since we reuse the link twice I was thinking about pulling it up somewhere but you're right, since James is working on it.

PR #55, opened 2024-12-13
Connect to mongo, and code for inserting movies

audiodude:
This is rescuing the Mongo code that was abandoned in #21


audiodude:
Note that this code is not called from anywhere, but I tested it in the repl.

PR #56, opened 2024-12-13
Provide script that downloads netflix prize data and puts it in the right place

### LINKED ISSUE #43: Provide script that downloads Netflix prize data and puts it in the right place
For our scripts to work, we need to download the Netflix prize data. As this file is ~500 MB and the licensing terms are dubious, we decided early on not to include it as part of the repo. It is however available on the Internet Archive here: https://archive.org/details/nf_prize_dataset.tar

We should write a script that downloads this tarfile, decompresses it, and puts the contents in `/data/`.

Specifically, everything that is in the `download` directory that the tarfile produces should be in the `/data/` directory. So if there is `download/movie_titles.txt` it should be `/data/movie_titles.txt`.

In general, this would be a good candidate for a bash script that does wget -> tar -xzf -> rename directory. But it would be better to do it in Python for portability reasons (some team members use Windows).

We can definitely download the file with `requests`, but it might not be the perfect fit. We should research if there is a `wget` type library for Python that we could use.

We can use the `tarfile` library to extract the files: https://docs.python.org/3/library/tarfile.html

Finally, renaming can be done with `os` or `shutil`: https://docs.python.org/3/library/os.html#os.rename

JamesKohlsRepo:
Fixes #43


audiodude:
Fixes #43

Pipfile
audiodude, line 1: This file needs to be merged with main. It should be simple, just keep everything that main added, and add your new dependencies and scripts.

Pipfile
```diff
@@ -5,13 +5,18 @@ name = "pypi"
 
 [packages]
 python-dotenv = "~=1.0"
-requests = "~=2.26"
 ruff = "~=0.7"
 tqdm = "~=4.66"
+requests = "*"
```
audiodude, line 10: We should leave the existing requests dependency.

Pipfile
```diff
@@ -5,13 +5,18 @@ name = "pypi"
...
 
 [dev-packages]
 pytest = "~=8.3"
 pytest-cov = "~=5.0"
 
 [requires]
 python_version = "3.12"
+
+[scripts]
+    download_data = "python download_data.py"
```
audiodude, line 21: These need to be `pipenv run python -m mediabridge.download_data`. The `pipenv run` is needed to make the scripts run in the pipenv environment.

mediabridge/clean_up_data.py
audiodude, line 1: I would recommend moving these scripts to the top level, not under the mediabridge module. The reasoning is that they are utilities for helping set up the environment, not a part of our software/library.

audiodude, line 1: This might be a bit overkill, because folks could just run `rm -rf data` to remove the data dir, and your download script will recreate it right?

mediabridge/download_data.py
```diff
@@ -0,0 +1,67 @@
+import os
+import tarfile
+
+import wget
+
+"""
+For our scripts to work, we need to download the Netflix prize data. As this file is ~500 MB and the licensing terms are dubious, we decided early on not to include it as part of the repo. It is however available on the Internet Archive here: https://archive.org/details/nf_prize_dataset.tar
```
audiodude, line 7: I think this comment only needs the first paragraph. The rest is a description of how to write this very script.

mediabridge/download_data.py
```diff
@@ -0,0 +1,67 @@
...
+    if not os.path.exists(target_file):
+        print("target file does not exist. Nothing to move")
+    else:
+        os.rename(target_file, destination)
+        print(f"Moving {filename}...")
+        print("Download Complete!")
+
+
+if __name__ == "__main__":
+    url = "https://archive.org/download/nf_prize_dataset.tar/nf_prize_dataset.tar.gz"
```
audiodude, line 59: Not necessary, but the usual pattern is to do:

```
def main():
   url = ...
   ...
   download_file(...)
   extract_file()

if __name__ == "__main__":
   main()
```

The main reason (see what I did there?) is that you might later add logic for handling command line args or other things, that are independent of the functionality of the `main()` function itself. Also, if you make `main()` its own function, it can be called from other scripts/modules (if that is ever useful).

audiodude (changes requested):
This looks great!

When you merge main, go ahead and delete the `Pipfile.lock` file. Then merge your changes to `Pipfile`, and run `pipenv install`. That will regenerate a fresh `Pipfile.lock`. The file is machine generated and not intended to be merged by hand.

Pipfile
```diff
@@ -5,13 +5,18 @@ name = "pypi"
 
 [packages]
 python-dotenv = "~=1.0"
-requests = "~=2.26"
 ruff = "~=0.7"
 tqdm = "~=4.66"
+requests = "*"
+wget = "*"
```
audiodude, line 11: Also, please specify a loose version requirement, like the other libraries in this file.

`wget = "~=9.9"`

Or whatever the current major/minor version is. You can get it off pypi, or with `pipenv run pip freeze`. Don't specify the patch version (so 9.9, not 9.9.9), so that we are pegged to the major version and the minor version can increment (versus being pegged to the minor version and only the patch version can increment). For more information see https://semver.org/

mediabridge/clean_up_data.py
audiodude: On second thought, it might be nice to keep them in `mediabridge` so that they can be imported from other scripts. We talked once about having an "uber" script that downloads the netflix data, does the matching, inserts into mongo and then does the recommendations. Having this in `mediabridge` would make that easier, though you might think about putting it in `data_processing`.

JamesKohlsRepo: I like the idea of keeping it in MediaBridge but relocating the files to data_processing

JamesKohlsRepo: I was thinking in the future it could be used to delete specific data, but it's probably a YAGNI situation

audiodude:
Hey @JamesKohlsRepo!

I was going to try and resolve your merge conflicts for you so we could get this PR merged. However, it looks like you made your changes in your own fork. In the future, it might be a good idea to create branches in this repo directly (you're a contributor) so that other folks can work on the code too.

jhanley634:
After merging down 72, this branch and PR should be deleted. Thx.

Issue #57, opened 2024-12-13
Flesh out MovieData class and use it for Mongo insertion


audiodude:
We have a MovieData dataclass where we store information taken from Wikidata. However, we also have information taken from the Netflix data set (netflix_id, title, year) that should be part of the class.

Once it's all in there, we can do `movie.asdict()` to convert the data class into a dictionary object, which can be put directly into Mongo.

Issue #34, opened 2024-10-27
Add text file to document contribution guidelines


skyfenton:
As we keep deciding on how tests should be constructed, naming schemes, formatting tools, docstrings, etc. we should have a markdown text file (a "CONTRIBUTING.md") to define our guidelines for writing code, submitting PRs, etc. that we can refer newcomers to and use to remind ourselves.

This way we also have a list of conventions to double check for in PR reviews to maintain consistency and confidence a PR has checked all the boxes.

jhanley634:
I can share things that I have seen are effective at "herding cats" when working with teams having between two and two dozen engineers of diverse skill levels.

- tests: `$ make test` is Green, and we have decent coverage. When adding a new feature function, expect that test code will exercise at least one line of that function. Or queue up a subsequent PR to add a test.
- naming: [PEP 8](https://peps.python.org/pep-0008/#naming-conventions). It's easy to do, and it's universal within the python community. Linters and reviewers can flag issues.
- formatting: [black](https://pypi.org/project/black), then verify a clean `$ ruff check`. One can get ruff to write fixes, but I find that makes it harder to communicate an idea to newbies than a simple "black + ruff", "write + read", would be. Even simpler: `$ make lint`, and have that [group imports](https://peps.python.org/pep-0008/#imports) with [isort](https://pypi.org/project/isort). I tend to assume that modern python code shall have type annotations -- no reason to omit them.
- docstrings: This is the only topic that is "hard".

We conduct a code review during the PR process to assess two things:
1. Is it correct?
2. Is it maintainable?

The easy-to-run `$ make lint test` gives an objective answer the first. I offer [this example](https://github.com/jhanley634/dojo-blackboard).

The author likely _believes_ that anyone in the team could fix a bug or add a feature to the code they just wrote. But only another teammate could give an answer of, "yes! I think I grok this new code and could maintain it." That's why we do reviews, to helpfully reveal blind spots in the author's thinking. What is obvious to the author won't be universally obvious, so we give other folks an opportunity to chime in, before they're stuck with supporting that code.

[...]

PR #61, opened 2025-01-17
Refactor main file/commands and expand dataclass usage

skyfenton:
Adds CLI commands for processing data into a csv and triggering a load of the csv into MongoDB.

Also creates dataclasses for input and enriched data (MovieData and EnrichedMovieData) and refactors code in wiki_to_netflix so that we can normalize the netflix data and write data into csvs based on these classes. Additionally, now when we write a list of MovieData objects to a csv file, the first line of the csv will consist of a header naming each column so that the entire file is mostly self-documenting (though, without explicit types). For example, movie_titles.csv writes EnrichedMovieData (matches), so the resulting csv file looks like:

```
netflix_id,title,year,wikidata_id,genres,director
3,Character,1997,Q928545,drama film; film based on literature,Mike van Diem
5,The Rise and Fall of ECW,2004,Q10381750,None,Kevin Dunn
...
```


mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -146,68 +147,61 @@ def format_sparql_query(title, year):
...
+    if not data["results"]["bindings"]:
+        log.warning(
+            f"Could not find movie id {movie.netflix_id}: (' {movie.title} ', {movie.year})"
+        )
+    else:
+        log.info(
+            f"Found movie id {movie.netflix_id}: (' {movie.title} ', {movie.year})"
+        )
+        return EnrichedMovieData(
+            **movie.__dict__,
```
audiodude, line 198: I think we want to use the dataclass method `asdict` right?

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -237,50 +231,21 @@ def process_data(num_rows=None, output_missing_csv_path=None):
...
-                title,
-                year,
-                "null",
-                "null",
-            ]
-            missing.append(movie)
+    netflix_data = read_netflix_txt(movie_data_path, num_rows)
+    for row in netflix_data:
+        id, year, title = row
+        netflix_data = MovieData(int(id), title, int(year))
```
audiodude, line 237: I don't think we need to repurpose the id or year as ints, because we're not doing any math with them right? Definitely not the id, I guess in some world we would want to query `year > 2000` or something.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -237,50 +231,21 @@ def process_data(num_rows=None, output_missing_csv_path=None):
...
+        if wiki_data := wiki_query(netflix_data, USER_AGENT):
```
skyfenton, line 238: There are a few ways to check if there is a match and instantiate a working MovieData object, but I figured it would be more readable if wiki_query returned either a new EnrichedMovieData object if it finds a match or None if it doesn't (rather than passing the netflix row itself and returning a MovieData or EnrichedMovieData object). Could change?

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -292,9 +257,47 @@ def process_data(num_rows=None, output_missing_csv_path=None):
...
+            raise e
+
+
 if __name__ == "__main__":
-    # Test is true if no argument is passed or if the first argument is not '--prod'.
-    test = len(sys.argv) < 2 or sys.argv[1] != "--prod"
-    process_data(
-        num_rows=DEFAULT_TEST_ROWS if test else None,
-    )
+    app()
```
audiodude, line 303: Nice!

mediabridge/dataclasses.py
```diff
@@ -0,0 +1,36 @@
...
+
+    netflix_id: int
+    title: str
+    year: int
+
+    def flatten_values(self):
+        """Format all dataclass fields into a 1-d list of strings by joining
+        lists with semicolons"""
+        return [
+            "; ".join(val) if type(val) is list else str(val)
```
audiodude, line 17: Prefer `val instanceof list` because it handles subclasses.

audiodude: Probably want to drop the space from the separator. The genres could have spaces in them, and using just a semicolon will make them easier to parse back in in the future.

mediabridge/dataclasses.py
```diff
@@ -0,0 +1,36 @@
...
+            for val in self.__dict__.values()
```
audiodude, line 18: As hinted by the fact that it's a double underscore, you probably want to avoid `__dict__`. I think you want `asdict` as well here.

mediabridge/dataclasses.py
```diff
@@ -0,0 +1,36 @@
+from dataclasses import dataclass
+from typing import Optional
+
+
+@dataclass(order=True)
+class MovieData:
+    """Dataclass for known data from the Netflix dataset"""
+
+    netflix_id: int
```
audiodude, line 9: As mentioned before, I don't think the id needs to be an int

mediabridge/dataclasses.py
```diff
@@ -0,0 +1,36 @@
...
+@dataclass(order=True)
+class EnrichedMovieData(MovieData):
+    """Dataclass for enriched data from a Wikidata match"""
+
+    wikidata_id: str
+    genres: Optional[list[str]]
+    director: Optional[str]
+
+
+if __name__ == "__main__":
```
audiodude, line 31: Is this like a fake test lol? Maybe you should write a real test, that takes a EnrichedMovieData and asserts that `flatten_values` returns the right thing.

mediabridge/db/load.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+@app.command()
+def load():
+    """
+    Load a csv of movie data into the mongo database.
+    """
+    pass
+
+    # db = connect_to_mongo()
```
audiodude, line 13: nit: no need to keep commented code, unless there's a TODO to include it in the future.

mediabridge/db/queries.py
```diff
@@ -5,9 +5,9 @@ def insert_into_mongo(movie):
     db = connect_to_mongo()
     collection = db["movies"]
     collection.update_one(
-        {"wikidata_id": movie[1]},
+        {"_id": movie[1]},
```
audiodude, line 8: The Mongo `_id` is the primary key that Mongo uses for indexing all the data. If we leave it blank, it is autogenerated by Mongo.

We probably need to have a larger discussion about what "id spaces" we're working with. So far we have 3: the netflix id, the wikidata id, and a third kind of "hidden/implicit" id that we might assign to the movies for reference in our own system. If we let Mongo assign database ids, then we're letting it handle that third scenario.

All I'm saying is that whatever is in the `_id` field is basically the "Mediabridge ID" of the movie, and we should think intentionally about how we want to assign/manage those.

mediabridge/main.py
```diff
@@ -1,42 +1,32 @@
...
 
 import typer as typer
-from tqdm.contrib.logging import logging_redirect_tqdm
 
 from mediabridge.data_processing import wiki_to_netflix
+from mediabridge.db import load
 from mediabridge.definitions import OUTPUT_DIR
 
+app = typer.Typer(no_args_is_help=True, add_completion=False)
+app.add_typer(wiki_to_netflix.app)
```
audiodude, line 12: Very cool!

mediabridge/db/queries.py
```diff
@@ -5,9 +5,9 @@ def insert_into_mongo(movie):
...
         {
-            "set": {
+            "$set": {
```
audiodude, line 10: Sorry, didn't we say that we wanted to use the `EnrichedMovieData` directly in the Mongo assignment? So this should just be:

```
"$set": movie.asdict()
```

right?

audiodude (changes requested):
This is great! I think that the ideas here are really good and they put us in a good place to run the code and manage the data.

I think all of the issues I've pointed out are minor and should be easily fixable.

audiodude:
Also @JamesKohlsRepo , please do leave your own review! I think learning how to review code is a really important skill to have.

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Yes, I completely agree that this is a good pattern: :+1:

mediabridge/db/load.py
skyfenton: Kept it since I didn't know if I was going to use it for implementing the load function, but just implemented it so we all good.

mediabridge/data_processing/wiki_to_netflix.py
skyfenton: `asdict` performs a full deep copy so it's a lot slower. In comparison, __dict__ just accesses the object's attributes dictionary directly.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -292,9 +257,47 @@ def process_data(num_rows=None, output_missing_csv_path=None):
...
+    # But when logging to file, we use nullcontext or tqdm will redirect logs
+    # back to stdout.
+    with nullcontext() if ctx.obj and ctx.obj.log else logging_redirect_tqdm():
+        num_rows = None if full else num_rows
+        try:
+            process_data(num_rows, output_missing_csv_path=missing_out_path)
+        except Exception as e:
+            # include fatal exceptions with traceback in logs
+            if ctx.obj and ctx.obj.log:
+                logging.exception("Uncaught exception")
```
jhanley634, line 298: Schulyer, thanks, there's lots of great code and refactoring in there.

I have just one substantive review item -- the remainder are all readability nit picking, feel free to ignore them, it won't hurt my feelings.

In `process()`, this message isn't very diagnostic and won't help a maintenance engineer find a fix a bug:
```
            if ctx.obj and ctx.obj.log:
                logging.exception("Uncaught exception")
```
Consider either removing it, or logging what `e` is about. OTOH the re-raise is great!

----

nit: Consider doing `from typer import Typer`, so the app assignment is more easily read.

I know you didn't edit it, but consider asking your IDE to rename
WikidataServiceTimeoutException -->
WikidataServiceTimeoutError, following convention.
(Two syllables is slightly easier to pronounce.)


```
def create_netflix_csv(csv_path, data_list: list[MovieData]):
```
Thank you for the beautiful list annotation.
Might as well point out we accept `csv_path: Path` and we return `) -> None:`

Producing headers might be slightly more convenient in this way:
```
writer = csv.DictWriter(netflix_csv, fieldnames=dataclasses.fields(data_list[0]))
writer.writeheader()
writer.writerows( ... )
```

In `wiki_query` it would be convenient for the caller if we defaulted `user_agent`.

The idiomatic `.raise_for_status()` looks great.

Consider removing the `else` clause, and finishing with
unconditional `.warning("Could not find..."); return None`.
The idea is those two go together in the execution flow,
so they should also go together on the printed page of source code.

In `process_data()` (and in definitions.py), instead of `.joinpath()`
consider using the idiomatic slash: 
```
output_csv = OUTPUT_DIR / "movie_titles.csv"
```

In `process()`, maybe the mixed  
`--num_rows` is fine? But consider using the conventional spelling of
`--num-rows`, which is easier for humans to type at the CLI. Similarly for
`--missing-out-path`. I think both are recognized, and this spelling just changes the --help output.

As an aid to legibility, consider adding "redundant" parens:
```
    with nullcontext() if (ctx.obj and ctx.obj.log) else logging_redirect_tqdm():
```

I feel this PR is definitely ready to emerge from Draft status.

jhanley634 (approved):
Travis is the one to keep happy, not me! But my tuppence is: LGTM :shipit:

mediabridge/data_processing/wiki_to_netflix.py
skyfenton: This conversion is a little useless because we only actually use each field as strings, but it's just here for some basic type checking to make sure the fields line up (there's a number in the year field, not a string, etc.). It would be better to do some more validation to make sure, say, the year is four digits, but I'm just pushing that down the line.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -237,50 +232,23 @@ def process_data(num_rows=None, output_missing_csv_path=None):
...
-                netflix_id,
-                "null",
-                title,
-                year,
-                "null",
-                "null",
-            ]
-            missing.append(movie)
+    netflix_data = read_netflix_txt(movie_data_path, num_rows)
+    for row in tqdm(netflix_data, total=num_rows):
```
skyfenton, line 238: I consolidated the loop to iterate on the read_netflix_txt generator instead of loading all the csv rows into a list, so tqdm needs to know the number of rows we ask for to show a progress bar. Only regression from this is we don't get a progress bar when we run it on the entire file because we don't know how long the file will be until we get to the end. I'm of the opinion this is fine?

skyfenton: To be fair, this is a pretty unnecessary optimization right now because 17000 lines is not that much anyway, but I was just cleaning up while refactoring. Am I missing a benefit to loading the entire file before parsing?

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Honestly, we don't really need to validate this data, because it's coming from the netflix set, and that never changes or updates. So if it's all valid years today, it will be all valid years forever.

The ID really isn't an int though right? Sometimes folks store IDs in databases as ints so they will be faster and take up less space, but that doesn't apply to us because we're storing in Mongo as JSON.

Anyways, I'm fine to leave as is.

mediabridge/data_processing/wiki_to_netflix.py
audiodude: You've got it exactly right. The only benefit to loading the entire file is to know how many rows there are ahead of time, so we can do things like display progress and (maybe, but we're not doing it) allocate space or resources or something.

In almost all data processing, it is preferred to iterate over and process data items one by one than to load everything ahead of time. So this is a welcome improvement!

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -146,68 +148,61 @@ def format_sparql_query(title, year):
...
 
     Parameters:
-    data_csv (list of lists): Rows of movie data with [movie ID, release year, title].
-    user_agent (str): used to identify our script when sending requests to Wikidata SPARQL API.
+    movie (MovieData): A MovieData object to use in the sparql query.
+    user_agent (str): Used to identify our script when sending requests to Wikidata SPARQL API.
 
     Returns:
-        list of WikiMovieData: A list of movieData instances with movie IDs, genres, and directors.
+    EnrichedMovieData: An EnrichedMovieData object containing information about the movie.
```
audiodude, line 160: I completely forgot to mention this, but the fact that `wiki_query` was actually `do_all_the_wiki_queries` and returned an entire list, has bugged me for some time. Thanks for fixing this!

mediabridge/dataclasses.py
skyfenton: You're right, though do you know if there are classes that inherit from list?

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Thanks John for the feedback!

To be clear, I'm not trying to "gatekeep" this project, or insert myself in the critical path as a reviewer or approver or vetoer or anything. These are just my honest code review comments. Like I said, I agree that most of them are minor.

> I have just one substantive review item -- the remainder are all readability nit picking, feel free to ignore them, it won't hurt my feelings.
> 
> In `process()`, this message isn't very diagnostic and won't help a maintenance engineer find a fix a bug:
> 
> ```
>             if ctx.obj and ctx.obj.log:
>                 logging.exception("Uncaught exception")
> ```
> 
> Consider either removing it, or logging what `e` is about. OTOH the re-raise is great!
> 

+1

> nit: Consider doing `from typer import Typer`, so the app assignment is more easily read.

I prefer to keep it as is, unless there is a way that I'm not remembering to do `from typer import Typer` but still keep everything else as `typer.Context`. I think that is more readable than just `Context` because it's not clear what that is.

> In `wiki_query` it would be convenient for the caller if we defaulted `user_agent`.

+1, the user_agent never changes.

> Consider removing the `else` clause, and finishing with unconditional `.warning("Could not find..."); return None`. The idea is those two go together in the execution flow, so they should also go together on the printed page of source code.

+1. Reverse the if (if <we have data>: return the data), and then since you returned in the if, you can just have the log and return None after that.

> In `process_data()` (and in definitions.py), instead of `.joinpath()` consider using the idiomatic slash:
> 
> ```
> output_csv = OUTPUT_DIR / "movie_titles.csv"
> ```
> 
> In `process()`, maybe the mixed `--num_rows` is fine? But consider using the conventional spelling of `--num-rows`, which is easier for humans to type at the CLI. Similarly for `--missing-out-path`. I think both are recognized, and this spelling just changes the --help output.

Disagree. The argument is indicated with `--` so it's error prone and confusing to separate words with a slash. All UNIX tools I can think of use --arguments_like_this=foo

mediabridge/dataclasses.py
skyfenton: It is lol, I was just lazy and didn't want to figure how to make the file structure readable for dataclasses and their tests. 

I could shove it into a directory called dataclasses with some tests, but then the import would become `from mediabridge.dataclasses.dataclasses import MovieData`. Instead we could rename the file in the folder to be ex. 'movies.py' so the import is more descriptive ex. `from mediabridge.dataclasses.movies import MovieData`, but if we add other files for dataclasses, then every dataclass' implementation and tests live together in the same directory (movies, movies_test, movies_test_data, games, games_test, games_test_data, ...) instead of inside their own subdirectories, but that's probably alright?

mediabridge/dataclasses.py
skyfenton: More details earlier, but we also just want to use __dict__ here so we can directly access all the attribute values (objects or not) of the instance and convert them to strings via join or str.

mediabridge/dataclasses.py
skyfenton: Replied on other comment (used just for simple type validation which we can expand in future)

mediabridge/dataclasses.py
skyfenton: Just asked chatgpt if it knew of an example, for anyone interested:

https://docs.python.org/3/library/collections.html#collections.UserList

mediabridge/data_processing/wiki_to_netflix.py
skyfenton: Cool cool. In the case of the netflix dataset, the id is always a number. Also I just do the conversions to prevent stupidity, like if on line 2409 of the dataset we accidentally insert an 'a' into the year field and we go debugging the failing SPARQL query because it's still just a string.

mediabridge/db/queries.py
skyfenton: Yeah, I was going to mess with it but I figured I'd let James handle this in [#57](https://github.com/noisebridge/MediaBridge/issues/57) since asdict might be the correct choice here to turn everything into dictionaries, even if it is more expensive. I'm not 100% sure, though, there might be some function in the json package to handle this.

mediabridge/db/queries.py
skyfenton: Very true, and I think we started the talk but didn't quite run into when it'd be used yet. I chose wiki_data.id here while testing because we are only uploading matches which all have unique wikidata ids and it would be easier for us to correlate visually, but maybe a better solution is a primary, mongo-generated "mediabridge" index and a secondary "wikidata" index (which I assume is possible) so we're not directly basing the main index off a db outside our system. Worth discussing in [#57](https://github.com/noisebridge/MediaBridge/issues/57).

jhanley634:
Hullo, @skyfenton. Here's a tiny feature request. In wiki_query(), please delete the surrounding pair of SPACE characters.
```
f"Could not find movie id {movie.netflix_id}: (' {movie.title} ', {movie.year})"
```
I thought we needed to add a title .strip() call, based on what the log was telling me, but it turns out it's just an output artifact that should be removed.

BTW, here's a pair of handy f-string syntax tips that might be of interest:
```
f"Hello {name=}"
f"Goodbye {name:r}"
```
The first is great for quick debugging -- it gives the identifier followed by its value.
In the second, instead of the usual `str()` we call `repr()`, which is boring for integers but it will put quotes around a string.

----

Also, `mypy` complains the process() signature isn't quite right.
We should declare `num_rows: int | None = typer.Option( ...`, given that we sometimes assign `None` to it.

jhanley634:
Ok, I finally found a substantive issue.

This PR 61 introduces a new `dataclasses.py` module. Please don't do that. At a minimum, rename it to `data_classes.py`. It causes confusion, and type checking lossage for mypy and pyright, which I eventually noticed and diagnosed.

Consider `$ python -c 'import test'`. It succeeds silently, due to one of python's standard libraries. A great many people have run afoul of that, naively creating their own test.py and then wondering why import behaves oddly. Same thing here.

mediabridge/data_processing/wiki_to_netflix.py
audiodude: I guess, but there's no reason anyone would insert an 'a' into the data right? I can't think of any reason we would need to edit those files, or would be even attempting to. Like I said, it doesn't matter.

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Did you actually benchmark it? We're talking about what, 4-5 strings? There aren't even any objects in the dataclasss, so "deep" copy isn't relevant. The __dict__ property of an object represents every property that the object contains, which could contain private or surprising data, which is why it is double underscored. For example, if we added a `to_json` method to the dataclas, that method itself would be part of `__dict__`, which would lead to bugs in this situation. The proper use of a dataclass is to use `asdict`, that's what the method is provided for.

mediabridge/dataclasses.py
audiodude: I wasn't sure off the top of my head if there were any built-in or standard library classes that inherit from list, but it's completely possible that we could use a class from an external library or something that does.

In any case, idiomatic python is `instanceof`.

mediabridge/dataclasses.py
audiodude: I think this is what you want: https://docs.python.org/3/library/dataclasses.html#dataclasses.fields

You should only use double underscore properties in very niche, specific, and extraordinary circumstances.

mediabridge/dataclasses.py
audiodude: I think `dataclasses_test.py` is fine.

You can also do what I do with my models in another project, and have an `__init__.py` initialization script in the module that "re-exports" (as it's called) the classes. https://github.com/audiodude/rainfall/blob/main/rainfall/models/__init__.py

So here we have `rainfall.models.artwork` which defines the `Artwork` class. But because we import it in `__init__.py`, we can do `from rainfall.models import Artwork`

mediabridge/db/queries.py
audiodude: Yes, I think you've got it right. We want to keep the external identifiers, but provide our own ID space, since something with a Mediabridge ID might exist in multiple, or none, of the external sources.

Yes you can add additional indexes to Mongo.

We do need to store "unenriched" movies though, because we need to be able to map between movie name -> netflix id and back to movie name.

mediabridge/db/queries.py
audiodude: It really isn't "expensive". Could you link the Stack Overflow answer that said that so I can go tell that person they're wrong?

audiodude (commented):
Thanks for the changes, looking good!

Point of order, we generally want to do:

1. Reviewer leaves comment
2. Author pushes updated code (optional)
3. Author says on comment: "Okay I changed that" or "No, I don't think it should change"
4. Reviewer checks that the new thing is good, or the explanation on why it doesn't need to change makes sense, and resolves comment.

I guess all that should be in #34

audiodude:
> This PR 61 introduces a new `dataclasses.py` module. Please don't do that. At a minimum, rename it to `data_classes.py`. It causes confusion, and type checking lossage for mypy and pyright, which I eventually noticed and diagnosed.

Great catch, John!

I would prefer to call it "models" or "data_models".

mediabridge/data_processing/wiki_to_netflix.py
skyfenton: Also saying thank you for the feedback! The more I get, the more experience I can say I have during interviews haha.

> In process(), this message isn't very diagnostic and won't help a maintenance engineer find a fix a bug:
> 
>             if ctx.obj and ctx.obj.log:
>                 logging.exception("Uncaught exception")
> 
> Consider either removing it, or logging what e is about. OTOH the re-raise is great!

Actually, logging.exception prints a traceback with the original exception details; this section is to write the exception to the log file if the (now renamed) log_to_file flag is true, then re-raise the exception so the actual program gets interrupted. To "unhide" this functionality so that it's clear, I changed the call in a recent commit to explicitly pass exc_info=True to logging.exception, even though it is True by default.


> In process(), maybe the mixed
> --num_rows is fine? But consider using the conventional spelling of
> --num-rows, which is easier for humans to type at the CLI. Similarly for
> --missing-out-path. I think both are recognized, and this spelling just changes the --help output.

Looks like dashes are the most correct and intuitive way based on the [gnu.org](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html) pages, and I agree with your reasons, so I committed a change for it.


> In process_data() (and in definitions.py), instead of .joinpath()
> consider using the idiomatic slash:
> 
> output_csv = OUTPUT_DIR / "movie_titles.csv"

It's funny, I just found out about this like two days ago while working on my own project. I like how it reads so I changed it, even though I feel a little icky not explicitly calling anything.


All the other nits I addressed in commits except the typer import because I think the typer.Option constructor keeps it clear Option is coming from Typer. :)

audiodude: > Disagree. The argument is indicated with `--` so it's error prone and confusing to separate words with a dash. All UNIX tools I can think of use --arguments_like_this=foo

Wow you're right @skyfenton! I don't know where I've been for the past 30 years. That GNU page is correct, and looking at man pages for `grep` and `cat`, they all do --long-options-like-this=whatever. Apologies.

skyfenton:
> > This PR 61 introduces a new `dataclasses.py` module. Please don't do that. At a minimum, rename it to `data_classes.py`. It causes confusion, and type checking lossage for mypy and pyright, which I eventually noticed and diagnosed.
> 
> Great catch, John!
> 
> I would prefer to call it "models" or "data_models".

How about we rename the "models" directory to "recommender" and then use the "models" directory to store the "dataclasses.py" file as "movie.py" or "moviedata.py" with files for "movie_test.py" and "moviedata_test_data.py"? I don't know if there's a better name other than just the base related dataclass name (MovieData). We could also move it into a "schemas" directory instead.

audiodude:
> How about we rename the "models" directory to "recommender" and then use the "models" directory to store the "dataclasses.py" file as "movie.py" or "moviedata.py" with files for "movie_test.py" and "moviedata_test_data.py"? I don't know if there's a better name other than just the base related dataclass name (MovieData). We could also move it into a "schemas" directory instead.

Oh crap, I forgot we had a directory for models already! Good catch. I think that was meant for models in the ML sense. 

Great ideas all around. No need for more bike shedding, the important part is to drop the name `dataclasses` for the module since it conflicts with the built in module. Do what you think is best.

mediabridge/data_processing/wiki_to_netflix.py
skyfenton: We would be using asdict every time we want to create a new instance which extends another instance's attributes, which is going to be very often since, for example, every time we match a movie to an entry in Wikidata (which we hope to get to be all 17770 movies in the list), we would be using asdict to copy the values to a new EnrichedMovieData instance. I benchmarked it earlier today, though, by running the system on the full dataset, but changing the wiki_query code to immediately return a new instance of EnrichedMovieData based on the given MovieData's attributes in these three ways:

```python
    # vars(object) (I found out this is the correct way to get a dict of an object's attributes)
    return EnrichedMovieData(
        **vars(movie),
        wikidata_id="TEST",
        genres=["TEST"],
        director="TEST",
    )

    # asdict
    return EnrichedMovieData(
        **dataclasses.asdict(movie),
        wikidata_id="TEST",
        genres=["TEST"],
        director="TEST",
    )

    # direct reference
    return EnrichedMovieData(
        netflix_id=movie.netflix_id,
        title=movie.title,
        year=movie.year,
        wikidata_id="TEST",
        genres=["TEST"],
        director="TEST",
    )
```

Granted, I just ran them a few times each at different points in time, I don't know how accurate tqdm is, and my CPU could be bursting or running processes in the background, all very unofficial, but tqdm showed pretty consistently that direct reference performed at a (extrapolated based on 17770 items) rate of ~310k items/s, vars() ~290k items/s, and asdict ~220k items/s. All imperceptible right now since we're not processing 300k items, but I think the overhead of doing deep copies of the attribute values is important, especially if we end up adding nested objects to these schemas, which asdict would recurse into/deep copy where direct reference and vars() would not copy at all.

mediabridge/dataclasses.py
skyfenton: You're totally right, I found out vars(obj) is the call I want to be making to access an objects attributes : values as a dictionary.

mediabridge/db/queries.py
skyfenton: The enriched movies would still have the Netflix id data, just more fields for the extra information we gather from wikidata. Maybe there's a better way of storing this we can figure out under the other issue.

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Oh I hadn't heard of `vars()`, but that seems much more appropriate that `__dict__`. Sounds good.

mediabridge/db/queries.py
audiodude: Okay I think this is good for now.

audiodude (approved):
Thanks for all your work updating this PR! I think the only thing left is testing, but I guess we've already set the (unfortunate) precedent that we don't write tests, so this is fine as is.

skyfenton:
> Also, mypy complains the process() signature isn't quite right.
> We should declare num_rows: int | None = typer.Option( ..., given that we sometimes assign None to it.

Fixed in recent change, maybe we should setup mypy or some other type checker (or maybe at least know about errors even if we don't want them to block)? 


> This PR 61 introduces a new `dataclasses.py` module. Please don't do that. At a minimum, rename it to `data_classes.py`. It causes confusion, and type checking lossage for mypy and pyright, which I eventually noticed and diagnosed.

Good catch! Renamed dataclasses.py to schemas.py instead, which I think is fitting, even more so if we eventually use pydantic.

Also removed spaces from outputs with titles and wrapped with double quotes.

skyfenton:
> Thanks for all your work updating this PR! I think the only thing left is testing, but I guess we've already set the (unfortunate) precedent that we don't write tests, so this is fine as is.

I added a couple tests for the flatten_values and wiki_query functions using the new dataclasses if you want to take a look. I discovered the order we get the genres of a movie is non-deterministic, so I sort the list before checking for equality.

skyfenton:
By the way, I'm gonna try to polish the cli a little more so I'll publish the pr once I'm done.

audiodude:
Sounds good! Nice work.

jhanley634:
Hmm, this is slightly sad. I'm accustomed to "trust the author!" settings. Oh, well, I'll just reapprove.

>  Review required
>
> Waiting on 1 reapproval from someone other than the last pusher. Reviews from audiodude and jhanley634 are stale because they were submitted before the most recent code changes. 

> Merging is blocked
> Merging can be performed automatically with 1 approving review.

mediabridge/schemas/movies_test.py
```diff
@@ -0,0 +1,15 @@
...
+        1, "The Matrix", 1999, "Q11424", ["Action", "Drama"], "Lana Wachowski"
+    ).flatten_values()
+    assert vals == {
+        "netflix_id": "1",
+        "wikidata_id": "Q11424",
+        "title": "The Matrix",
+        "year": "1999",
+        "genres": "Action;Drama",
+        "director": "Lana Wachowski",
+    }
```
jhanley634, line 15: LGTM, :shipit: !

jhanley634 (approved):
re-approving...

audiodude:
I wonder if there is a way to require review/approval, but not re-approval?

jhanley634:
> I wonder if there is a way to require review/approval, but not re-approval?

Yup. Here are two items. And for background, my personal philosophy is that Authors are smart and good, they mean well, the Author is always right, can always choose to ignore a Reviewer's remark. My role as Reviewer is simply to offer a different perspective on the code that perhaps was never considered. Once a PR has a 👍 thumbs up on it, I wish for Author to be able to quickly e.g. fix a typo and merge to `main`. We assume giant changes wouldn't piggyback on the initial approval -- they should go in a subsequent feature branch.

I usually click github repo Settings to allow this on Pull Requests:
- [no] Allow merge commits Loading 
- [yes] Allow squash merging
- [no] Allow rebase merging

Then everyone who clicks the giant green Merge button gets a Squash, so `git annotate` history is simplified. I tend to do at least a commit an hour, often with tiny changes, and they will be uninteresting a month from now. Showing the granularity of "introduced feature X" in the log makes it more informative, more easily understood.

To answer your original question, it's on branch protection Rules. Make sure that "Dismiss stale pull request approvals when new commits are pushed" is disabled. Which ensures that previous approvals won't be invalidated when new commits are added to the branch.

audiodude:
Okay yup, I found the rule, thanks!

PR #66, opened 2025-01-21
Recommendation engine class

cocomittens:
Ok this is currently untested & still working on it, but I believe at least the converting ids to titles functionalities should theoretically work? Just putting this up here to show progress.


mediabridge/engine/RecommendationEngine.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_movie_title(self, netflix_id):
+        movies = self.db["movies"]
+        return movies.find_one({"netflix_id": netflix_id}).get("title")
+
+    def titles_to_ids(self, titles):
+        netflix_ids = []
+        for title in titles:
+            netflix_id = self.get_movie_id(title)
+            titles.append(netflix_id)
+        return netflix_ids
```
jhanley634, line 29: FYI, if we call this approach (A.), there's a pair of identical ways to accomplish the same thing.

(B.)
def titles_to_ids(self, titles):
    return [self.get_movie_id(title) for title in titles]

(C.)
def titles_to_ids(self, titles):
    return list(map(self.get_movie_id, titles))

Is one of these "better"? No, not exactly. During early development and debugging we might prefer (A.), as we can insert debug print()'s, breakpoint(), and so on. Later on a list comprehension might make more sense. I personally prefer (C.) because it highlights how very very simple the processing is -- we're just applying a  function, nothing else, no tricks up my sleeves, it's just one function. But YMMV.

Notice that generators like map() and range() are "lazy", so we use list() to pull all the results out of the generator. Consider range(2) versus list(range(2)).

(Please advance this out of "draft" status once the code is ready for merging to `main`.)

jhanley634:
Please advance this out of "draft" status once the code is ready for merging to `main`.

jhanley634 (approved):
LGTM, :shipit: !

PR #69, opened 2025-01-22
67 measure coverage

jhanley634:


jhanley634:
This sets us up for figuring out what code and what tests are important, or are missing.
Run `bin/coverage.sh` and then visit `htmlcov/index.html` in your favorite web browser.

There are some entirely unrelated mediabridge-frontend/package-lock.json edits, due to dependabot diagnostics I received on each push. I would be happy to abandon / revert such edits here, and let someone deal with PR #68 much better than I could. Even after adopting what appeared to be the recommended vite version, I kept seeing those push diagnostics.

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -1,26 +1,31 @@
...
+        QUERY = w_to_n.format_sparql_query("The Room", 2003)
+        assert QUERY == EXPECTED_SPARQL_QUERY
 
-def test_wiki_query():
-    movie = MovieData("0", "The Room", 2003)
-    result = w_to_n.wiki_query(movie)
-    assert result
+    def test_wiki_query(self) -> None:
+        movie = MovieData("0", "The Room", 2003)
+        result = w_to_n.wiki_query(movie)
```
audiodude, line 15: This does an actual query to Wikidata correct?

We should prefer so-called "hermetic" tests for unit tests: https://medium.com/geekculture/hermetic-unit-testing-8c49276e3acd

If we wish to test actual Wikidata responses, it should be done in an integration/e2e test.

audiodude (approved):
If you can, would you mind setting up pipenv "scripts" for your shell scripts in bin? That way we can do`pipenv run coverage`, the same way we already have pipenv scripts set up for running other code.

You're correct that it wasn't necessary to fix the dependabot issues in this PR. Dependabot issues will only register as "resolved" once the updated dependency code is pushed to `main`. So really we could have just merged the dependabot PR to fix. Generally we want to keep PRs focused to more or less one issue.

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: Yes, that's correct. I added a comment to clarify.

In order, things I don't like are:
- red (failing) test
- flaky test (I confess that wikidata network errors could potentially impact this test, though I haven't seen it)
- missing test
- slow test (I confess that network latency could make this painfully slow, though I haven't seen it)

I was just using an existing test already in the codebase. I feel that test's author made a sensible "unit vs integration" tradeoff, writing a test for the existing Public API exported by wiki_query(). At this point we would need to alter that API and pass in a sparql_result_getter rather than a query_endpoint string. Or we could set up a Flask listener on localhost which produces a canned result. All of which could be good, but it's out-of-scope for the current PR.

The **real** leverage point is for each Author to **write the test first**, and then write the target code. This immediately surfaces "it's hard to test" concerns, as the code is fluid and evolving, and tends to affect what Public API we choose to expose. Once we have good coverage, I hope that will provide encouragement to folks who are inventing new routines and wondering how to convince themselves, and other project participants, that a new routine is working properly.

PR #71, opened 2025-01-23
Make wiki_to_netflix_test hermetic

audiodude:
Unit tests should be hermetic. They should not make any network requests. They should not depend on an external API.

This is for multiple reasons:

1. Network calls are inherently unreliable. Making network calls in unit tests will make them flaky. This might not be apparently when only doing one or two, but when a test suite grows to 1000 tests, it will get increasingly harder to run the test suite successfully
2. Tests which require network calls also require that developers who wish to work on the project be online, which is not always possible or desirable. Hermetic unit tests mean that the code can be worked on without a network connection.
3. We may wish to run our unit tests in CI environments that we do not control. Therefore, we are not able to guarantee the composition or topology of the network in those environments.

It is definitely the case that we want to know when, for instance, the Wikidata API changes in a way that breaks our code. If Wikidata starts returning extra fields that break us, or omits data we need, etc, we should respond accordingly. However, this should not be discovered in _unit_ tests. This is the subject for separate, integration/end-to-end tests. The unit tests should verify that given a certain input (including input from an external API), the code under test performs in an expected way.

### DIFF
--- a/mediabridge/data_processing/wiki_to_netflix_test.py
+++ b/mediabridge/data_processing/wiki_to_netflix_test.py
@@ -1,17 +1,22 @@
 import unittest
+from unittest.mock import patch
 
 import mediabridge.data_processing.wiki_to_netflix as w_to_n
-from mediabridge.data_processing.wiki_to_netflix_test_data import EXPECTED_SPARQL_QUERY
+from mediabridge.data_processing.wiki_to_netflix_test_data import (
+    EXPECTED_SPARQL_QUERY,
+    WIKIDATA_RESPONSE_THE_ROOM,
+)
 from mediabridge.schemas.movies import EnrichedMovieData, MovieData
 
 
 class TestWikiToNetflix(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


skyfenton (approved):
LGTM!

I was going to do something like this, but I didn't know how we should construct tests for cases in which the API changed, so I just picked the simplest option for the moment (which isn't the smartest idea, of course).

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -217,6 +217,7 @@ def wiki_query(
 
     response.raise_for_status()
     data = response.json()
+    print(repr(data))
```
jhanley634, line 220: Leftover debug?

jhanley634 (approved):

mediabridge/data_processing/wiki_to_netflix.py
audiodude: Yup! Removed.

PR #72, opened 2025-01-24
Download data

JamesKohlsRepo:
Converted my fork into a branch called Download-Data


Pipfile
```diff
@@ -5,13 +5,18 @@ name = "pypi"
 
 [packages]
 python-dotenv = "~=1.0"
-requests = "~=2.26"
 ruff = "~=0.7"
 tqdm = "~=4.66"
+requests = "~=2.32"
+wget = "~=3.2"
```
jhanley634, line 11: nit: Prefer to keep the "requests" entry where it was please, in alphabetic order. In general, if "order does not matter" in some source file, then we prefer sorted order so it's easy for maintenance engineers to see where a new entry should go, reducing merge conflicts. On which topic...

The github website UI explains that this branch has conflicts with `main` and so is not yet ready to merge. If you want help with that, just ask and one of us can step in with some edits on this branch. But using `$ git mergetool` is a skill worth learning, and there's no time like the present, so you might as well start now. Remember that `$ git status` is your friend. Run it often, to stay aware of current status, and to read the helpful advice it sometimes offers, such as "run git commit to finish merging the changes from main". Note that Pipfile.lock is not _exactly_ like other source files, as it is derived, so a blind "regenerate and accept Our changes" suffices for that one.

I think `main` has some additional [scripts] entries which do not yet appear in this branch.

I use `opendiff` from Xcode, FYI.
```
$ cat ~/.gitconfig
[user]
    name = John Hanley
    email = jhanley634@users.noreply.github.com
[alias]
    co = checkout
    st = status
[credential]
    helper = osxkeychain
[merge]
    tool = opendiff
```

mediabridge/data_processing/download_data.py
```diff
@@ -0,0 +1,63 @@
...
+    source_path = os.path.join(download_dir, file_name)
+    destination_path = os.path.join(data_dir, file_name)
+
+    download_file(url, compressed_file_path)
+    extract_file(compressed_file_path, data_dir, movie_titles_path)
+    move_files(source_path, destination_path)
+
+
+if __name__ == "__main__":
+    main()
```
jhanley634, line 63: In clean_data_folder(), please attend to that "f-string with no {variables}" `ruff` diagnostic, preferring a simple string over an f-string. Ideally you would have run `ruff` prior to `$ git push`. Prefer to only push code which lints cleanly.

Delete the last two lines of wiki_to_netflix.py, as appending blank lines doesn't improve this PR.

The various function- and module- level docstrings are lovely and I thank you for them.

The comment `# create directory if not already existing` is obvious and says what the code already eloquently said; please delete it. Code reveals the HOW details of what to do, and comments should reveal the WHY.

Ok, here's a review remark that I feel you should **not** act upon in the current PR, but I would like you to consider the advice when you're writing code in future. The verbose os.path API is slightly painful to read at this point; prefer the more modern pathlib API. See e.g. the `x = DATA_DIR / y` format used [here](https://github.com/noisebridge/MediaBridge/blob/8c44d233/mediabridge/data_processing/wiki_to_netflix.py#L263). Please _do_ merge down the current code as-is, since I am merely making a "six versus half-dozen" remark here -- they both work fine. Just keep it in mind for what you write next week.

jhanley634 (approved):

jhanley634:
Hullo, James. Just checking in. This branch looks like it's nearly ready for a merge. Do my remarks mostly make sense? Is there any trouble with `$ git mergetool` for resolving merge conflicts against `main`? We are here to help -- pipe up and we can lend a hand.

mediabridge/data_processing/download_data.py
skyfenton: +1 for the pathlib library, joins and checking if paths exist are super easy.

Pipfile
JamesKohlsRepo: Thank you, I am addressing these issues

mediabridge/data_processing/download_data.py
JamesKohlsRepo: I appreciate you detailed recommendations, sorry for not being timely with my contributions

mediabridge/data_processing/clean_up_data.py
```diff
@@ -0,0 +1,34 @@
...
+            print(f"Deleted file: {item_path}")
+        elif os.path.isdir(item_path):
+            shutil.rmtree(item_path)
+            print(f"Deleted directory: {item_path}")
+
+    print("All contents of the 'data' folder have been deleted.")
+
+
+def main():
+    data_dir = os.path.join(os.path.dirname(__file__), "../../data")
```
jhanley634, line 29: Certainly this works, and you can merge it as-is. But consider calling [normpath()](https://docs.python.org/3/library/os.path.html#os.path.normpath) on it, for better quality diagnostics if / when something downstream goes south. The pathlib equivalent would be [.resolve()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve). The "../.." in the middle of what we're passing around has unambiguous meaning to the machine. But humans tend to get a little lost with all the up and down when trying to interpret a message like "bad thing happened with foo/bar/../baz/file.txt".

If you _do_ choose to revisit this prior to merging down, consider [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)ing up the two different definitions of `data_dir` so there's just one. The project's usual idiom seems to be: `from mediabridge.definitions import DATA_DIR`

(The clean_data_folder() seems like a long-winded way of saying .rmtree() to nuke followed by mkdir(), but maybe there's some permission attribute on the data/ directory that we're trying to preserve.)

----

It looks like you still need to get Pipfile merged, or ask one of us to jump in on your feature branch and merge it for you. Fortunately it's small, so it's easy to eyeball the differences with `main` branch, which you can conveniently view on the github site in a web browser. Keeping entries alphabetical eases that task.

Don't worry too much about the .lock file, as Pipfile is the important one. A `pipenv lock` will regenerate it and we can just "use ours" to resolve any merge nonsense. Good luck! Remember to click "delete" your feature branch once the squash merge is complete.

jhanley634: @JamesKohlsRepo , I would really like to see this merge down prior to PR 81. I feel that merging in the other order would unfairly subject you to more "bait and switch", more "keep chasing the accursed Pipfile merge conflicts".

If you want me to jump in on this feature branch to fix up Pipfile and merge to `main`, just say so.

JamesKohlsRepo: I did neglect to update clean_up_data, I will revisit with your recommendations. Travis also pointed that out to me, I can try and address updating with rmtree() and observing DRY. 

I thought I figured it out on Tuesday but I'm still having issues. if you don't mind @jhanley634 or @skyfenton  helping me grasp this I would appreciate it. I'll try and tackle the remaining issues first.

mediabridge/data_processing/download_data.py
```diff
@@ -0,0 +1,48 @@
...
+    data_dir = Path(__file__).resolve().parent.parent / "data"
+    download_dir = data_dir / "download"
+    compressed_file_path = data_dir / compressed_filename
+    file_name = "movie_titles.txt"
+    source_path = download_dir / file_name
+    destination_path = data_dir / file_name
+
+    download_file(url, compressed_file_path)
+    extract_file(compressed_file_path, data_dir)
+    move_files(source_path, destination_path)
```
skyfenton, line 44: You shouldn't need to move the files, just download the .tar.gz archive to the data directory, extract it into the current folder, then delete .tar.gz. I think we generally want the whole dataset.

skyfenton, line 44: You shouldn't need to move the files, just download the .tar.gz archive to the data directory, extract it into the current folder, then delete the .tar.gz. I think we generally want the whole dataset.

Pipfile
```diff
@@ -4,14 +4,32 @@ verify_ssl = true
...
 pytest-cov = "~=5.0"
+ruff = "~=0.7"
+types-requests = "~=2.32"
+types-tqdm = "~=4.67"
 
 [requires]
 python_version = "3.12"
+
+[scripts]
+download_data = "pipenv run python -m mediabridge.data_processing.download_data"
```
skyfenton, line 30: I'll fix this when I go back over the CLI, but I think this should be a 'mediabridge init' command, since it's crucial for the code to work. I'm treating these pipenv scripts mainly as "developer shortcuts", though it's pretty ambiguous.

skyfenton (changes requested):
Heya, just a couple extra comments to help get this finished! Basically it seems like just five(ish) changes (take as many commits as you need):
1. Modify the download_data operations to extract the tar.gz download to the same folder, delete the original file, then extract training_set.tar into training_set/. 
2. Simplify clean_up_data (honestly, probably could just remove it and use just rm).
3. Reorder pipfile alphabetically
4. Delete pipfile.lock and run pipenv install again to regenerate it.
5. Pull from main and resolve any merge conflicts.

mediabridge/data_processing/download_data.py
audiodude: #85

mediabridge/data_processing/clean_up_data.py
audiodude: #85

audiodude:
I've fixed the merge conflict with the following commands:

```
git checkout download-data
git merge main
code .
git add Pipfile
rm Pipfile.lock
pipenv install
git add Pipfile.lock
git commit
git push origin download-data
```

The `code .` command is where I opened `Pipfile` in VSCode and manually merged the changes.

Let's merge this now and work on

Issue #44, opened 2024-11-08
Pair programming of tests


audiodude:
We should schedule attempt to pair/group programming session where we write tests together. Our Thursday night meeting slot would be a good time for this.

This is not a bug, it's more of a task/reminder.

PR #73, opened 2025-01-24
70 improve code coverage

jhanley634:
Adds tests + measurement infrastructure, to ensure we are thoroughly testing the `main` branch target code. Each
```
env PIPENV_IGNORE_VIRTUALENVS=1 pipenv run coverage
```
measurement run takes two seconds. Hopefully authors proposing a merge of new target code will take a peek at how much of that code is exercised by automated tests.

```
Name                                                       Stmts   Miss  Cover
------------------------------------------------------------------------------
mediabridge/data_processing/wiki_to_netflix.py               124     13    90%
mediabridge/data_processing/wiki_to_netflix_test.py           87      0   100%
mediabridge/data_processing/wiki_to_netflix_test_data.py       2      0   100%
mediabridge/definitions.py                                     7      0   100%
mediabridge/schemas/movies.py                                 13      0   100%
mediabridge/schemas/movies_test.py                             4      0   100%
------------------------------------------------------------------------------
TOTAL                                                        237     13    95%
```


mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +65,78 @@ def test_wiki_query(self, mock_requests):
...
+    def _read_full_netflix_txt(self) -> bool:
+        movies = list(w_to_n.read_netflix_txt(TITLES_TXT))
+        assert len(movies) == 17_770
+        assert movies[-1] == ["17770", "2003", "Alien Hunter"]
+        return True
+
+    def test_create_netflix_csv(self) -> None:
+        output_csv = DATA_DIR / "movie_titles.csv"
+        few_rows = 3  # A conveniently small number of rows, for fast tests.
+        rows = list(w_to_n.read_netflix_txt(TITLES_TXT, few_rows))
```
audiodude, line 101: Tests should be independent of one another. You should not use the output of one library function as the input to another that is under test. The reason for this is that if one of the functions start failing, it will cause cascading failures throughout the other tests, which will be difficult to debug and reason about.

You are only testing 3 movies anyways, should you should just be able to:

```
movies = [
  MovieData(1, 'Dinosaur Plant', 1997),
  MovieData(2, 'The Thing', 1988),
...
]
```

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +65,78 @@ def test_wiki_query(self, mock_requests):
...
+        movies = [MovieData(id, title, int(year)) for id, year, title in rows]
+
+        w_to_n.create_netflix_csv(output_csv, movies)
+        lines = output_csv.read_text().splitlines()
+
+        assert len(lines) == few_rows + 1
+        assert lines[-1] == "3,Character,1997"
```
audiodude, line 108: nit: There are only 3 lines, why not assert all of them?

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +65,78 @@ def test_wiki_query(self, mock_requests):
...
+
+    def test_wiki_feature_info(self) -> None:
+        # This is a silly test, designed to provoke `return None`,
```
audiodude, line 111: Regarding this comment, although we curently don't have any configuration of SPARQL or input data that results in the bindings being missing, we don't control the API so we don't know what it will return in the future, and if we make any changes that result in empty output, it is useful to handle that case.

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +65,78 @@ def test_wiki_query(self, mock_requests):
...
+        # something that live wikidata results is never observed to do.
+        # (Don't believe me? Put an `assert False` in that function and see if it triggers.)
+        # Recommend deleting this test, and the target code clause it is written for.
+        no_entries: list[str] = []
+        data = {"results": {"bindings": no_entries}}
```
audiodude, line 116: nit: `no_entries` is just an empty list, why not inline it in data below?

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +65,78 @@ def test_wiki_query(self, mock_requests):
...
+    def _process_with_path_missing(self, actual: Path) -> None:
+        """Verifies we get an error from process_data(), when a path is momentarily is missing."""
+        ephemeral = Path(f"{actual}-missing")
+        assert actual.exists()
+        actual.rename(ephemeral)
+        try:
+            with self.assertRaises(FileNotFoundError):
+                w_to_n.process_data(1, None)
+        finally:
+            ephemeral.rename(actual)
```
audiodude, line 128: I don't understand the point of this part. You've sucessfully tested that the function raises an Exception if the path is missing. What does it matter that missing path then gets moved to the "actual" path?

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +65,78 @@ def test_wiki_query(self, mock_requests):
...
+
+    def test_process_data_pretend_theres_no_data_dir(self) -> None:
+        self._process_with_path_missing(DATA_DIR)
+        self._process_with_path_missing(TITLES_TXT)
+
+    def test_process(self) -> None:
```
audiodude, line 134: This is a test with no assertions, so technically, it doesn't test anything. It just runs the library code. Please update using assertions or remove the test.

audiodude (changes requested):
Thanks for writing all these test, we definitely have been missing them! Though I had hoped we could write them together at some Thursday session (see #44), but it's useful to have them anyways.

jhanley634:
There's a lot of details here. I think we're mostly on the same page. There are some test details that I should probably finesse prior to merging. Let's do a call tomorrow. I'm available around 11am, on 650-888-0110.

audiodude:
> There's a lot of details here. I think we're mostly on the same page. There are some test details that I should probably finesse prior to merging. Let's do a call tomorrow. I'm available around 11am, on 650-888-0110.

I don't think a call is necessary? If you have questions about my comments, feel free to ask in this PR and I will respond and clarify if possible.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -41,6 +58,7 @@ def read_netflix_txt(
     Yields:
         List of strings representing the values of the next row in the file.
     """
+    ensure_movie_titles_txt_exists(txt_file)
```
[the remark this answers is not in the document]
jhanley634: I am not yet sure what I should do with this condrum -- please offer more guidance.

I share your "ideally", but the project is not yet in an ideal position. The very concrete thing I was trying to do within the scope of **this** PR was to satisfy these two requirements:
1. coverage.sh on a developer's laptop shall read > 17,000 titles and shall report "high" coverage, e.g. 90%
2. pytest in the github Action CI environment shall report Success, despite the fact that 17,000 titles are not available in the CI environment.

The current PR satisfies both those requirements. What I want to better understand is, do you as Reviewer agree that both should be a requirement? And what alternate implementation are you suggesting that would continue to satisfy both requirements?

I mean, I _could_ just comment out the ensure_movie_titles_txt_exists() call or whatever. But then we would see pytest fail in CI. Are we maybe thinking about "cheating", where tests ask "what environment am I running in?" and suppress certain actions in CI?

I'm aware that James has been working on the "fetch movie info" problem for some months. I explicitly did not want that to be a blocking issue for this PR, nor for this PR to interfere with that work. I was trying to scope down the PR so we can get it out the door, leaving the codebase in better shape than it was. And then if the codebase improves, for example if `main` gets some "fetch movie info" function, we can always do a bit of refactoring. For example we might insist that every environment (developer, CI) shall always have 17,000 titles available. I just don't see how to enforce that given the **current** state of `main`.

mediabridge/data_processing/wiki_to_netflix.py
```diff
@@ -103,9 +121,8 @@ def wiki_feature_info(data: dict, key: str) -> str | list | None:
 def wiki_feature_optional_str(data: dict[str, Any], key: str) -> str | None:
     """Validates that we obtained a single (optional) string."""
     s = wiki_feature_info(data, key)
-    if s:
-        return str(s)
-    return None
+    assert isinstance(s, (str, NoneType)), s
```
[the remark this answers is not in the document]
jhanley634: I do not yet understand why this is a review concern. Is the proposed change somehow leaving the code in worse shape than it was before?

We could comment it out, which leads to this lint output:
```
wiki_to_netflix.py:124: error: Incompatible return value type (got "str | list[Any] | None", expected "str | None")  [return-value]
```
So the goal of that line of code is to convince `mypy` that only `str` and `None` could possibly be coming back. (I imagine we might see `director = None` movies, though I haven't encountered any yet.) I was trying to take that notion and implement it in code in the most obvious, direct way I could think of. This might a "six versus half-dozen" review remark, where there's more than one way to get a turing machine to do what you want, and neither the six approach nor the half-dozen approach is clearly better than the other.

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -1,21 +1,51 @@
...
     EXPECTED_SPARQL_QUERY,
     WIKIDATA_RESPONSE_THE_ROOM,
 )
+from mediabridge.definitions import DATA_DIR
 from mediabridge.schemas.movies import EnrichedMovieData, MovieData
 
+TITLES_TXT = DATA_DIR / "movie_titles.txt"
+
+
+def silence_logging(self, logger: Logger):
```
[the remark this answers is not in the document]
jhanley634: Yeah, that's a good call, thanks. Done.

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -1,21 +1,51 @@
...
+            yield  # Let the test code run for a bit.
+        finally:
+            logger.handlers = original_handlers  # Turn logging back on.
+
+    return _silence_logging()
+
 
 class TestWikiToNetflix(unittest.TestCase):
-    def test_format_sparql_query(self):
+    def test_format_sparql_query(self) -> None:
```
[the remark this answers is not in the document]
jhanley634: Philosophically I agree with you. But as a practical matter it's necessary, so we avoid this diagnostic:
```
wiki_to_netflix_test.py:22: error: Function is missing a return type annotation  [no-untyped-def]
wiki_to_netflix_test.py:22: note: Use "-> None" if function does not return a value
```
I find linting is more helpful when we use `mypy --strict`. That option is not currently enabled by default because only a subset of the codebase lints clean with it. This PR makes a handful of source files lint cleanly now.

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: Yeah, I initially considered asserting equality with a triple-quoted string. But I felt that was far too much overlap between the test and the target code. 
The essential concept I'm trying to get across is that the file is "small", it has fewer than 17,000 movies. I confess that the first len() assert suffices for that. The second string equality assert is the sort of "dirt simple" test I tend to habitually put into automated tests. It gives the "flavor" of the data we're working with, it's a reminder of what shape it has. It also verifies the file doesn't end with some spurious blank line. I would be happy to remove that second assert if you wish.

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: I _kind_ of agree. But another ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)) perspective is that it suffices for the target code to reliably throw fatal error when "surprising" input data arrives, and we could worry about the details then. One good outcome from coverage measurement is it can reveal which parts of the target code are "impossible" to execute, and sometimes it will motivate changing the Public API so individual failure modes are better modeled, such as by turning `validate_and_process()` into a pair of `validate()` and `process()` functions, or by making caller responsible for fetching a JSON dict via the internet so that `process(d: dict)` simply cannot fail with a network timeout.
In any event I didn't feel strongly about it, and wasn't going to start changing anyone's API. The comment was mostly a reminder to me about "perhaps in future we might be able to do away with this silly test altogether." One can always dream, and sometimes I write down such dreams in comments.

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: Sure, I understand, that's an entirely sensible question.
Suppose we just assign `no_entries = []`, which works the same as passing in `[]` as an argument. Well, in that case we have deprived the type checker of some essential type information. `[7, 8]` is clearly "list of int", and `["a", "b"]` is clearly "list of str". But in the empty case we can't distinguish those, which elicits this diagnostic:
```
wiki_to_netflix_test.py:94: error: Need type annotation for "no_entries" (hint: "no_entries: list[<type>] = ...")  [var-annotated]
```

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: I didn't feel entirely good about this, and you're right to question and perhaps insist I do it a bit differently. Also, the target code is maybe a bit weird. It is trying hard, perhaps too hard, to offer a good **diagnostic** error message. It first checks for `data/` directory, then checks for movie titles text file. Arguably it would be enough for the code to hope for the best and throw a fatal FileNotFound error, which an engineer (_probably_?) could quickly interpret as "oh, time to go fetch a few thousand movie titles from the net!". Alternatively, with code from James merged to `main`, it might be the case that that failure mode simply "never happens", so the code could be a little less paranoid. I chose to _not_ go in and alter the target code.

The aspect of this test that I wasn't entirely pleased with was the "dangerous" operation of moving a giant file of movie titles out of the way, asking target code to notice it doesn't exist, and then quickly move it back as though nothing had happened. My concern was that some unexpected bad thing might happen within the target code, and we never get around to moving the file back where it belongs. It just feels like a slightly ugly approach. But it's expedient, so I went with it.

It was belief at the time that various bits of code "knew" where the titles file lives, and I didn't want to mess with that. But perhaps a better approach is to parameterize its location, and then a trivial "please go open /non/existant.txt" test can provoke the error path. I'm certainly open to pursuing that if you want. It probably does not fit within the scope of the current PR.

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -1,21 +1,51 @@
...
 
 import mediabridge.data_processing.wiki_to_netflix as w_to_n
 from mediabridge.data_processing.wiki_to_netflix_test_data import (
     EXPECTED_SPARQL_QUERY,
     WIKIDATA_RESPONSE_THE_ROOM,
 )
+from mediabridge.definitions import DATA_DIR
 from mediabridge.schemas.movies import EnrichedMovieData, MovieData
 
+TITLES_TXT = DATA_DIR / "movie_titles.txt"
```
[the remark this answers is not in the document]
jhanley634: OIC, ok, let me code that up this afternoon.

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: Well, it runs the target code and verifies that no error is raised. But yeah, I agree, something from the target's contract should be verified. I think I just got too excited once I figured out how to put a well-formed context together so I could actually call the target, and then moved on.

Ok, we verify a result file was produced, now.

mediabridge/data_processing/wiki_to_netflix_test.py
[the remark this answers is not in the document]
jhanley634: I wound up having to change the Public API offered by process_data(), something I had been trying to avoid. But yeah, I agree shipping a tiny text file makes more sense.

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +65,78 @@ def test_wiki_query(self, mock_requests):
...
+
+    def test_read_netflix_txt(self) -> None:
+        movies = list(w_to_n.read_netflix_txt(TITLES_TXT, 3))
+        assert len(movies) == 3
+        assert movies[-1] == ["3", "1997", "Character"]
+
+        short = TITLES_TXT.stat().st_size <= 74
+        # Sometimes we're in an environment, like CI, where we never downloaded the full dataset.
+        # So silently succeed, without attempting to read thousands of non-existent entries.
+        assert short or self._read_full_netflix_txt()
```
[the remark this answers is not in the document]
jhanley634: Cool. Implemented.

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: Implemented.

jhanley634:
Ok @audiodude, I attended to a dozen review remarks. Please update this PR, either "approved" or describe any new edits you desire.

mediabridge/data_processing/wiki_to_netflix.py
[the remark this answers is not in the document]
audiodude: I completely agree that the tests should:

1. Run on a developer's computer successfully
2. Run in the CI environment successfully

However I don't think it's the case that the first scenario requires 17000 movies. Since the function already takes the path of the file as an input (dependency injection, perhaps), it is possible to provide it with a _test_ file in the _test_ environment (whether it be the developer's laptop or CI).

So what I would do is take the test data that you are writing in the `ensure_movie_titles_txt_exists` function, and write it to a permanent test data location. Then, in the test, simply pass that file.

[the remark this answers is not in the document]
jhanley634: @audiodude Ummm, I think maybe you're looking at an old copy of the branch, and you want to `$ git pull` ?
I nuked that `ensure_movie_titles_txt_exists()` function at your behest, and replaced it with a tiny text file. I agree with you that it's a cleaner approach.

mediabridge/data_processing/wiki_to_netflix.py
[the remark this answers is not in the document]
audiodude: Like I said, I haven't worked with mypy very much. I'm more used to Typescript, which has a similar philosophy of "type erasure", where the types are hints to the compiler (or in this case, the linter) but do not actually affect runtime code (or at least I think that's how Python types operate).

In Typescript, if a method returns a type that is "too wide" (aka list, str, or None), you can write a line of code that asserts that the type is not the one you don't want (in this case list), and the compiler will then let you access the "narrowed" type as they call it, without error.

That's what I was suggesting:

```
if s instance of list:
  raise ValueError("List is not valid here (you can probably write a better message)")
```

And I assumed that mypy would then be happy, because when you go to `return s` it can't possibly be list anymore right?

mediabridge/data_processing/wiki_to_netflix_test.py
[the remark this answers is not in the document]
audiodude: Okay that's a perfectly good reason to leave them then.

mediabridge/data_processing/wiki_to_netflix_test.py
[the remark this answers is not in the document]
audiodude: I think that testing often exposes weaknesses in API design. An API that is not easily testable is not a good API IMO. This change looks good!

mediabridge/data_processing/wiki_to_netflix.py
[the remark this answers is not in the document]
jhanley634: I completely agree with you about what the Turing machine would do -- we could assert "not a list". But I respectfully disagree that's the ideal way to express the notion in the source code. We write code primarily to convey a technical idea to other humans, and only secondarily to express an idea to the machine. The first part, clearly communicating to humans, is hard, and communicating an idea to the machine is relatively easy.
So rather than telling a collaborator "it's not a list!", I was focused on telling them "it's either this or that". Humans have cognitive limitations, and I feel that expressing something in the positive, rather than the negative, will better assist them in properly reasoning about what the code does.

mediabridge/data_processing/wiki_to_netflix_test.py
audiodude: I guess to me, it makes sense that if you want to generate a CSV, and you know what movies you're starting with, then you know exactly what the CSV should contain, and it's not "cheating" to assert that. To me, a good test sets up the input, runs the function under test, and then asserts the output. As you said, there aren't 17000 movies to assert against, so it makes sense to me to just assert all of them. But this is fine.

mediabridge/data_processing/wiki_to_netflix_test.py
audiodude: I agree, we could architect the function differently. It depends on what we consider a "fatal" error. My opinion is that if you're processing 17,000 movies, you should not die halfway through because a single one prompted an API response that was not what you expected. But we can have that discussion later.

I guess I wouldn't call a test that exercises valid code in a function to be "silly".

mediabridge/data_processing/wiki_to_netflix_test.py
audiodude: I disagree that we should be altering the code to jump through type checking hoops, but this is fine for now and I appreciate the explanation.

mediabridge/data_processing/wiki_to_netflix_test.py
audiodude: I understand what you're saying. I think that you're really just exposing the warts of the code under test, in that it has paths hardcoded into the function implementations that can not be parameterized. I think that the right thing is to, in fact, make the data locations parameters to these functions, but I also appreciate that you were just trying to write the tests without changing the code too much. We should file an issue to follow up.

mediabridge/data_processing/wiki_to_netflix_test.py
jhanley634: Duly noted.
Comment elided!

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -35,3 +45,76 @@ def test_wiki_query(self, mock_requests):
...
+
+        assert not TITLES_CSV.exists()
+
+    def test_process(self) -> None:
+        ctx = Context(command=example_command, info_name="mediabridge")
+        csv = FULL_TITLES_TXT.with_suffix(".csv")
+
+        if FULL_TITLES_TXT.exists():
+            # At this point we're on a developer's laptop, rather than in CI.
+            csv.unlink(missing_ok=True)
```
audiodude, line 110: I don't think we should erase files on a developer's machine when they run the test, unless it was a file that the test itself created. Please fix.

mediabridge/data_processing/wiki_to_netflix_test.py
audiodude: See comment about erasing developer work.

mediabridge/data_processing/wiki_to_netflix.py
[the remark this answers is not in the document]
audiodude: Sorry, I was just going through comments one by one and responding. Since you didn't say in this comment that you had done that, I didn't think to look at the code. My bad.

mediabridge/data_processing/wiki_to_netflix.py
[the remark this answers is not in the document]
audiodude: I agree with you, though maybe (if you're so keen on communicating with humans) you can add a comment along the lines of:

```
# The above function can sometimes return a list, which is not appropriate here
```

So that when I get a list and it errors out, I don't have to go digging through type signatures.

PR #77, opened 2025-01-27
76 enable   strict mode

jhanley634:
"A small PR is a beautiful PR."
This one has limited scope, and touches only a handful of files, making it easier to approve. It enables `mypy --strict`, so that authors may be reminded to annotate the signatures of newly added functions.
No functional changes. We are merely changing the type hints.

### DIFF
--- a/bin/lint.sh
+++ b/bin/lint.sh
@@ -7,6 +7,6 @@ source bin/util.sh
 
 ${ENV} ruff check .
 
-${ENV} mypy ${STRICT} --show-traceback --warn-unreachable --ignore-missing-imports --no-namespace-packages .
+${ENV} mypy --strict --show-traceback --warn-unreachable --ignore-missing-imports --no-namespace-packages .
 
 ${ENV} pyright .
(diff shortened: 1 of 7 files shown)


skyfenton (dismissed):
All LGTM if we want strict typing, which I'm in favor of. Maybe we should add a github actions job (or replace the current ruff check) to run lint.sh as well?

EDIT: I thought the request changes review action would be less serious, I meant to start a conversation to be resolved but didn't need to comment on code lol.

audiodude (approved):

jhanley634:
Hullo, @skyfenton, thanks for your input. I respectfully disagree.

Each author is free to lint before committing, and if that works for them, great! But I'm reluctant to impose that on them. Plus, type annotation isn't always easy, there are some odd cases which occasionally arise. If some code gets merged down without annotation, and a subsequent PR makes it lint clean, then it's all good, we can all get along within a single codebase.

Minimally the group should gain experience and day-to-day familiarity with annotating newly written code before diving all the way in. It's essentially a related, slightly different language that you're coding in, when following modern python conventions. Each person can get used to it in their own time.

Issue #88, opened 2025-02-15
engine coverage


jhanley634:
We need to exercise the RecommendationEngine code in a repeatable, reproducible way, which works the same for everyone. Pytest is one way to accomplish that. Grep'ing the git log for "Bugfix:" will reveal several misunderstandings that arose from different people seeing different things in the codebase. A standard test infrastructure serves to reduce such misunderstandings.

Assignments of the form `xs = []` are troublesome to type checkers, as it is unclear if we mean a list of int, or list of str, or something else. We _could_ annotate such assignments, but that impresses me as being a bit intrusive. Usually it's enough to just annotate signatures. So refactoring as `return [some list comprehension]' seems indicated. This is a refactor, and we'd like to see a Green test before and after a refactor. This motivates improving the code coverage, prior to adding type hinting.

Currently we see this:
```
pipenv run lint | egrep '/engine/|^Found'
+ env PYTHONPATH=.:../.. mypy --strict --show-traceback --warn-unreachable --ignore-missing-imports --no-namespace-packages .
mediabridge/engine/recommendation_engine.py:10: error: Function is missing a type annotation  [no-untyped-def]
mediabridge/engine/recommendation_engine.py:16: error: Function is missing a type annotation  [no-untyped-def]
mediabridge/engine/recommendation_engine.py:20: error: Function is missing a type annotation  [no-untyped-def]
mediabridge/engine/recommendation_engine.py:24: error: Function is missing a type annotation  [no-untyped-def]
mediabridge/engine/recommendation_engine.py:25: error: Need type annotation for "netflix_ids" (hint: "netflix_ids: list[<type>] = ...")  [var-annotated]
mediabridge/engine/recommendation_engine.py:27: error: Call to untyped function "get_movie_id" in typed context  [no-untyped-call]
mediabridge/engine/recommendation_engine.py:31: error: Function is missing a type annotation  [no-untyped-def]
mediabridge/engine/recommendation_engine.py:34: error: Call to untyped function "get_movie_title" in typed context  [no-untyped-call]
mediabridge/engine/recommendation_engine.py:38: error: Function is missing a type annotation  [no-untyped-def]
mediabridge/engine/recommendation_engine.py:46: error: Function is missing a return type annotation  [no-untyped-def]
mediabridge/engine/recommendation_engine.py:50: error: Function is missing a type annotation  [no-untyped-def]
[...]

Issue #83, opened 2025-02-07
engine annotations


jhanley634:
Add optional type annotations so RecommendationEngine will lint clean. Currently there's 18 lines of noise, some of it related to empty container assignments having ambiguous type. (Is `[]` a list of str? A list of int? Hard to say, until we hint it.)

This wound up turning into a mini-epic:
- 83-engine-hints, which contains no hints due to distraction of ETL speedup and discovering pymongo issues by running the code
- 88-engine-coverage, which exercises much of the engine code and discovered some difficulties
- 83-part2-engine-hints, which will use the coverage to safely refactor / hint the engine code
- future 83-part3-download-hints, maybe? (Trying to defer work on those files till after 90 merges.)

----

@skyfenton 
(Ok, I'm timing out on PR 90 that was approved Feb. 20th. It's Feb 23rd now. I had been deliberately holding back to avoid merge conflicts as 90 touches a very large number of files. I _do_ want to see some new edits merge down by Thursday.)

PR #89, opened 2025-02-15
88 engine coverage

jhanley634:
Improves test coverage, and fixes at least one bug.
There's some motivation in issue https://github.com/noisebridge/MediaBridge/issues/88

(PR #83 has been reviewed / approved / merged, so now reviewing this one will is _much_ simpler.)

----

This PR

1. Pickles a model in addition to a matrix; in the course of that the `output/` folder is abolished, in favor of putting everything in `out/`. Please use OUTPUT_DIR when writing new code.
2. Tells etl_mongo_movie_titles() to rely on read_netflix_txt(). Sorry -- merge conflict, PR 86 could have gone smoother.
3. Uses conventional terminology for mongo objects: connect_to_mongo() returns a `db`, not a `client`.
4. Adds import_lightfm_silently(), so stderr isn't cluttered with crazy warnings we're not in a position to fix.
5. Bugfix for a copy-pasta whoops from PR 66: titles_to_ids() wants to append a result id rather than changing the input titles param.
6. Adds test_get_movie_title, test_ids_to_titles, test_get_data, test_display_recommendations, test_create_user_matrix, test_create_user_matrix, and changes the `recommend` function so tests can obtain recommendation results from it.


tests/engine/recommendation_engine_test.py
```diff
@@ -20,18 +20,24 @@ def etl_mongo_movie_titles() -> None:
     movies_collection.insert_many(rows)
 
 
+# In a github Actions CI environment, we will be "disabled".
+_enabled = bool(os.getenv("MONGODB_URI"))
```
audiodude, line 17: We can set up CI so that Mongo is available.

mediabridge/data_processing/etl.py
```diff
@@ -1,23 +1,17 @@
...
 
-from mediabridge.data_processing.wiki_to_netflix import read_netflix_txt
-from mediabridge.db.tables import (
-    POPULAR_MOVIE_QUERY,
-    PROLIFIC_USER_QUERY,
-    Rating,
-    get_engine,
+from mediabridge.data_processing.etl_helpers import (
+    RATING_CSV,
+    _insert_ratings,
```
audiodude, line 9: I thought the convention in Python is that `_underscore_methods` are "private", to the extent that they should only be used in the module or class they are defined in.

tests/engine/recommendation_engine_test.py
jhanley634: Sure, that would be a good Issue to open. Sounds like it is out-of-scope for the present PR.

mediabridge/data_processing/etl.py
jhanley634: Quite right, point taken. (This is actually PR 86 code being reviewed in PR 89.)
I used to say "it's OK for a TestCase to reach in and exercise private _foo()", but nowadays I will typically promote it to public foo() just so it fits in smoothly with other tooling.
Here, my motivation is "how can I make the tools work better for me?" That is, I want `pipenv run coverage` to offer easily interpreted figures that convince me we're still running / testing the deployed code. So I was trying to banish "code that takes eight minutes to run" to another module in order to simplify the interpretation of a three-second interactive run. Possibly a conditional `import` would do that better.
In any event I won't merge these symbols down in PR 86 as-is. I will probably strip the leading underscore to make them public. The `ruff` linter lacks a config option to flag this due to common cases like `import socket` for accessing pure python while `import _socket` accesses the C language bindings.

jhanley634: Ok, there's much more sensible code over in PR 86 now. Once that hits `main` I will merge it into PR 89.

jhanley634:
@cocomittens , if you have example code which calls `.recommend()`, I'd be very happy to include it in this PR. Thx.

skyfenton (changes requested):
The code itself looks good I think, but I can't run the tests because line 60 of interaction_matrix.py puts the output file under a "/output" directory instead of using the OUTPUT_DIR definition. I think we just didn't catch it when we merged in changes a few weeks ago, but could you fix this for this PR so the tests can be evaluated?

jhanley634:
@skyfenton Yikes! Good catch. I will attend to it.

PR #90, opened 2025-02-18
Add CLI entrypoints to manage data download and loading

### LINKED ISSUE #85: Follow ups for download data scripts
Taken from #72.

Minor issues that we want to keep in mind for the future, or a follow up PR.

> I did neglect to update clean_up_data, I will revisit with your recommendations. Travis also pointed out removing the directory, I can try and address updating with rmtree() and observing DRY. 
> I thought I figured it out on Tuesday but I'm still having issues. if you don't mind @jhanley634 or @skyfenton  helping me grasp this I would appreciate it. I'll try and tackle the remaining issues first.

_Originally posted by @JamesKohlsRepo in https://github.com/noisebridge/MediaBridge/pull/72#discussion_r1945614358_

> You shouldn't need to move the files, just download the .tar.gz archive to the data directory, extract it into the current folder, then delete the .tar.gz. I think we generally want the whole dataset.

_Originally posted by @skyfenton in https://github.com/noisebridge/MediaBridge/pull/72#discussion_r1950034326_

>  I'll fix this when I go back over the CLI, but I think this should be a 'mediabridge init' command, since it's crucial for the code to work. I'm treating these pipenv scripts mainly as "developer shortcuts", though it's pretty ambiguous.

_Originally posted by @skyfenton in https://github.com/noisebridge/MediaBridge/pull/72#discussion_r1950038820_

skyfenton:
This PR unifies the interface for performing actions using mediabridge by the following:
- 'init' command to download the dataset in the appropriate directory
- 'load' command to load dataset data into the db tables for processing (with option to recreate all data with confirmation)
- 'clean' to delete the /data and /out dirs (with confirmation)

This PR also fixes some outstanding issues:
- all etl code now uses a path to the netflix dataset, NETFLIX_DATA_DIR (a subdir in /data)
- etl.py now creates tables if they are missing on first/regenerative runs
- downloader now correctly downloads all files to a new data/nf_prize_dataset directory, fixes #85 
- leftover tar files are now unlinked after download
- downloader now uses the requests library since the wget progress bar could interfere with logging

(Ignore @jhanley634's commits, I pulled in changes from another PR early and they ended up in this one)


mediabridge/data_download.py
```diff
@@ -0,0 +1,61 @@
...
+we decided early on not to include it as part of the repo. It is however available on the Internet Archive here: 
+https://archive.org/details/nf_prize_dataset.tar
+"""
+
+log = logging.getLogger(__name__)
+
+
+def download_file(url: str, output_path: Path) -> None:
+    output_path.parent.mkdir(parents=True, exist_ok=True)
+    r = requests.get(url, stream=True)
```
jhanley634, line 22: nit: It is conventional to do `r.raise_for_status()` here.
In the event of, say, 404, a subsequent call would fail. But raise_for_status offers a much better diagnostic, at the point of the error.

(When I say "nit:", I mean the code clearly works and could be merged down as-is. Consider an edit for this PR or when writing code next week. In particular, nits are not a GitHub "change requested" which would block merging to `main`.)

mediabridge/data_download.py
```diff
@@ -0,0 +1,61 @@
...
+        ):
+            if chunk:
+                f.write(chunk)
+                f.flush()
+
+
+def extract_file(src: Path, dest: Path, gz_compressed: bool = False) -> None:
+    print("\nExtracting...")
+    mode = "r:gz" if gz_compressed else "r:"
+    with tarfile.open(name=src, mode=mode) as tar:
```
jhanley634, line 37: nit: Managing the tedious details of the read `mode` seems more complex than necessary, as the library pays attention to the magic number prefix byes and automagically does the Right Thing. We are needlessly burdening the caller. Recall that the man page offers this example, with no `mode` passed in:
```
tar = tarfile.open("sample.tar.gz")
tar.extractall(filter='data')
```

mediabridge/data_download.py
```diff
@@ -0,0 +1,61 @@
...
+        tar.extractall(path=dest)
+
+
+def download_netflix_dataset() -> None:
+    NETFLIX_DATA_DIR.mkdir(parents=True)
+    url = "https://archive.org/download/nf_prize_dataset.tar/nf_prize_dataset.tar.gz"
+    compressed_filename = "nf_prize_dataset.tar.gz"
+    compressed_file_path = DATA_DIR / compressed_filename
```
jhanley634, line 45: nit: While `url` is not exactly a local filesystem Path, it is convenient to make it one. Because then we could assign `... = DATA_DIR / url.name`

mediabridge/data_processing/etl.py
```diff
@@ -14,16 +15,19 @@
...
+
+log = logging.getLogger(__name__)
 
 RATING_CSV = OUTPUT_DIR / "rating.csv"  # uncompressed, to accommodate sqlite .import
 
 GLOB = "mv_00*.txt"
 
 
-def etl(max_rows: int) -> None:
+def etl(max_reviews: int, regen: bool = False) -> None:
```
jhanley634, line 30: tiny nit: consider forcing caller to name the parameter, e.g. `etl(12, regen=True)`.
```
def etl(max_reviews: int, *, regen: bool = False) -> None:
```
We'll often do this for boolean parameters, as they can be a bit cryptic at the call site. This especially matters when we're passing in two or three booleans.

mediabridge/data_processing/etl.py
```diff
@@ -33,33 +37,46 @@ def etl(max_rows: int) -> None:
...
-        columns = ["id", "year", "title"]
-        df = pd.DataFrame(read_netflix_txt(FULL_TITLES_TXT), columns=columns)
-        df["year"] = df.year.replace("NULL", pd.NA).astype("Int16")
-        # At this point there's a df.id value of "1". Maybe it should be "00001"?
+    # if there is already data in movie_title, skip reprocessing
+    if not pd.read_sql_query(query, get_engine()).empty:
+        log.warning(
+            "movie_title table already populated with data, skipping reprocessing..."
+        )
+        return
```
jhanley634, line 66: nit: I confess I don't understand why this is a "warning". It feels like it would be pretty routine during a three-second edit-run cycle, where the table _was_ populated when pytest started, and will _remain_ populated afterward. The .warn feels too chatty.

mediabridge/data_processing/etl.py
```diff
@@ -33,33 +37,46 @@ def etl(max_rows: int) -> None:
...
+        df.to_sql("movie_title", conn, index=False, if_exists="append")
 
 
-def _etl_user_rating(max_rows: int) -> None:
-    """Writes out/rating.csv if needed, then populates rating table from it."""
-    training_folder = PROJECT_DIR.parent / "Netflix-Dataset/training_set/training_set"
-    diagnostic = "Please clone  https://github.com/deesethu/Netflix-Dataset.git"
-    assert training_folder.exists(), diagnostic
+def _etl_user_rating(max_reviews: int) -> None:
+    """Writes out/rating.csv.gz if needed, then populates rating table from it."""
```
jhanley634, line 78: Sorry, I don't understand why the docstring moved from .txt to .txt.gz. The sqlite .import command sadly requires an uncompressed file, so we never produce a .gz version. Please fix the docstring to match what the code does.
(Thank you for fixing my `_insert_ratings()` docstring, which once was true when it mentioned compressed. There is always the danger that "comments lie!".)

mediabridge/data_processing/wiki_to_netflix_test.py
```diff
@@ -53,29 +54,29 @@ def test_wiki_query_not_found(self) -> None:
...
+    #     movies = [
+    #         MovieData("1", year=2003, title="Dinosaur Planet"),
+    #         MovieData("8806", year=1966, title="The Good, the Bad, and the Ugly"),
+    #         MovieData("3", year=1997, title="Character"),
+    #     ]
+    #     w_to_n.create_netflix_csv(TITLES_CSV, movies)
+    #     lines = TITLES_CSV.read_text().splitlines()
+
+    #     assert len(lines) == len(movies) + 1
+    #     assert lines[-1] == "3,Character,1997"
```
jhanley634, line 79: As discussed in discord, I feel these `#` commented code lines should not be part of the code base. Rename the function, or simply delete the lines leaving behind a single SHA1 commit hash comment for folks that want to find them in the git history.

mediabridge/main.py
```diff
@@ -30,12 +35,14 @@ def main(
...
         print(
-            f"[WARNING] Output directory does not exist, creating new directory at {OUTPUT_DIR}"
+            "[WARNING] Output directory does not exist, "
+            f"creating new directory at {OUTPUT_DIR}"
         )
         OUTPUT_DIR.mkdir()
 
     if log:
         # log all messages to new file
+        # DOESN'T WORK IF REFRESHING OUTPUT DIRECTORY
```
jhanley634, line 45: Hmmm, this seems slightly worrying. I confess I don't know what exactly lies behind the "doesn't work" remark. Possible in-scope cleanup before merging this PR?

mediabridge/main.py
```diff
@@ -52,5 +59,41 @@ def main(
...
+    refresh: bool = typer.Option(
+        False, help="Deletes data-holding directories before initializing."
+    ),
+) -> None:
+    """Download all required datasets and initialize the database"""
+    if refresh:
+        # prompt = f"\n! This will delete all data in {DATA_DIR} and {OUTPUT_DIR}. Would you like to proceed? y/n !\n"
+        # if input(prompt) != "y":
+        #     print("\nAborting process.")
+        #     return
```
jhanley634, line 73: Please delete `#` commented code prior to merging down to `main`.

mediabridge/recommender/make_recommendation.py
```diff
@@ -137,7 +141,7 @@ def _get_test_movie_ids(
     """
     params = {"max_training_user_id": large_movie_id}
     with get_engine().connect() as conn:
-        test_movie_ids = [int(m) for m, in conn.execute(text(query), params)]
+        test_movie_ids = [int(m) for (m,) in conn.execute(text(query), params)]
```
jhanley634, line 144: Nit: this seems a needlessly verbose way to (talk) (about) (a) one-tuple.
We are, of course, unpacking the row's single element into a scalar, which `int` can work with.

jhanley634 (approved):
Mostly looks good. There's definitely some big improvements in here.
I impose no "requested changes" barrier, and trust the Author to exercise judgement and do appropriate things prior to merging to `main`.
Make zero or more edits that respond to the review remarks, squash, and delete feature branch, as your schedule permits.

jhanley634:
`mediabridge/data_download.py` is maybe not the _best_ module name. Certainly it's not terrible. But unlike the very generic `definitions.py`, it feels like it deserves to be pushed down a level into a subdirectory.

mediabridge/data_download.py
audiodude: Disagree. URLs are not paths just because they are separated by `/`.

jhanley634: Fair enough. That's what the "not exactly" waffling was about. Also, better to lodge such a complaint a week ago. This PR has languished without activity for far too long.

It's a fact that there's a relationship between URLs "out there" in the cloud, and Paths "local" to our host's filesystem. Especially for the final component. It helps with [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). We can conveniently model that.

```
class URL(Path):
    """Models FTP and similar pathnames, such as HTTPS filenames, which may later appear with same name in the local filesystem."""

def download_netflix_dataset() -> None: ...
    url = URL("https://archive.org/download ...")
    compressed_filename = url.name
    ...
```

mediabridge/data_processing/etl.py
audiodude: #95

mediabridge/main.py
skyfenton: I was using it as a reminder to implement some changes before publishing, but it's fixed now (added a directory for logs).

mediabridge/data_processing/wiki_to_netflix_test.py
skyfenton: Created issue #96 to address the integration tests, which includes the git hash where it was commented out. Deleted these comments in recent commit.

mediabridge/data_download.py
skyfenton: Path classes have a lot of methods that don't make sense for URLs (mkdir? anchor?), and in this case (and most cases in our project) we don't need to do any annoying URL manipulation, plus I'm not convinced pathlib would parse urls correctly. If we decide we need something more robust in the future, we should consider [yarl](https://pypi.org/project/yarl/).

mediabridge/data_download.py
skyfenton: Good catch! Added.

mediabridge/data_processing/etl.py
skyfenton: I don't know what happened since I swore I merged everything in main, but I corrected it in a recent commit.

mediabridge/recommender/make_recommendation.py
skyfenton: The ruff formatter does this by default if you leave a straggling comma (though it doesn't give an explicit warning).

jhanley634: Ooohhhh! Hmmm, interesting. It happens I have a small disagreement with `ruff` in this instance, but fine, whatever.
I recommend leaving the `(m,)` that ruff put there; it's not hurting anything.
I'm afraid I routinely rely on `$ black *.py` to put source code in canonical form, which `ruff check` will accept without complaint. This is because I find `black` does much, much better at dealing with irregular horizontal whitespace and vertical whitespace than `ruff check --fix` does. But it's all good, just merge as-is.

mediabridge/data_download.py
skyfenton: Added in 1b0cc77545b87a0fa8f6660f040ec456ee4e216a

mediabridge/data_processing/etl.py
skyfenton: Fixed in 6f809658b47846ec86b810022a78a1646ecbe5ac

PR #91, opened 2025-02-23
83 part2 engine hints

jhanley634:
This PR touches **two** python source files, in order to
1. add @beartype dynamic (runtime) type checking assertion
2. add type hints to RecommendationEngine methods, for static type checking with mypy + pyright
3. minor "use the catch_warnings context manager already!" cleanup
4. fix a "wrong type for liked_movies_ids" bug

(I am trying to avoid the "download" code for now, while PR 90 is still unresolved, but that's next on the todo list.)


jhanley634:
I deliberately made this PR small, so it's easy to approve. Anyone can weigh in here, to move things along. Typically a code review should complete within 24 hours, or at least should spark a conversation about what needs to be done. Thx.

skyfenton (commented):
All looks good, though I'm still hesitant about enforcing type hinting with bear. There are good arguments for and against it, and I would be for it on my own projects, but I don't know enough to know if the impact on newcomers/code checks is significant, ex. running into annoying cases where I have to go find which typing package supports which structures.

jhanley634:
There is a tool, like any other. Each tool has its strengths and weaknesses, it's sweet spot. When a tool sees to be helpful, then we get rid of it. What I encountered was that mypy and pyrite were not giving sufficiently helpful diagnostics to let me properly determine What type the original author of the target code intended. Those two are static analysis, tools, or as bare type is a dynamic analysis, tool, essentially an assert. If it gives what we feel are false negatives, then certainly we should disable it. If it measurably slows down execution of tight loops, then again we might consider disabling it in the name of CPU cycle efficiency. As it stands, I am unaware of any arguments against it, and I recommend merging it down into Maine. We can always revisit this decision in a subsequent PR.

jhanley634:
This is now in sync with `main`, post PR 90.

mediabridge/engine/recommendation_engine.py
```diff
@@ -1,78 +1,87 @@
...
         cols = liked_movies_ids
         data = [1] * len(liked_movies_ids)
-        return sparse.coo_matrix((data, (rows, cols)))
+        return coo_matrix((data, (rows, cols)))
 
-    def recommend(self, user_id: int = 0, limit: int = 10) -> set[int]:
+    def recommend(self, user_id: int, limit: int = 10) -> set[int]:
         liked_movies = self.get_data()
         liked_movies_ids = list(map(int, self.titles_to_ids(liked_movies)))
+        print(f"{liked_movies=}")
```
audiodude, line 81: Are these print statements supposed to be left in?

audiodude (approved):

mediabridge/engine/recommendation_engine.py
jhanley634: Ummm, IDK. It has been "a long time" since last I visited this source (Feb. 23rd at 12:29), so I don't exactly remember what my state of mind was. I think I agree with you, that this is "left over debug", and I will delete it prior to merging. I'm accustomed to PR's happening quickly enough that we still have the various concerns front-of-mind when considering alternative edits.

audiodude: We all know that PRs in this project aren't proceeding as fast as you would like, I'm not sure you need to keep bringing it up.

PR #93, opened 2025-02-24
92 download hints

jhanley634:
Hmmm. I thought this might be a bigger PR. But it appears that PR 90 has already dealt with much of the lint noise; that's a pleasant surprise!

(Do not attempt to interact with this PR 93 until the upstream PR 90 hits `main`.)


jhanley634:
[PR 91](https://github.com/noisebridge/MediaBridge/pull/91) should hit `main` before this gets merged. (It can be approved at any time.) Currently we see noise like this, which 91 should resolve:
```
mediabridge/engine/recommendation_engine.py:77: error: Call to untyped function "display_recommendations" in typed context  [no-untyped-call]
```

audiodude (approved):

PR #102, opened 2025-03-06
99 verify onboarding

jhanley634:
This PR adds containerized Dockerfile instructions, and fixes some "start from zero" bugs it surfaced.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+# build with:
+#
+#   time docker buildx build .
+
+FROM python:3.12-alpine
+
+WORKDIR /app
+
+COPY . .
+
+RUN apk add --no-cache gcc libffi-dev musl-dev \
+ && pip --root-user-action option install pipenv
+
+RUN pipenv lock \
+ && pipenv install --dev \
+ && pipenv run lint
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+COPY Pipfile .
+
+RUN sudo chown -R media:media /app \
+ && pipenv lock \
+ && pipenv install --dev
+
+COPY . .
+
+RUN sudo chown -R media:media /app \
+ && pipenv run coverage  > /tmp/coveage.log
```
audiodude, line 28: I don't think it's appropriate to run tests as part of building a Docker image. It should just install the app.

audiodude (changes requested):
Looks good, nice and simple. Just one question.

audiodude (approved):

Dockerfile
jhanley634: I agree. Will revise this week in a subsequent PR.

jhanley634: Fixed in a032572f: No need to run tests at the end of generating an image, as we're confident they shall work

Issue #103, opened 2025-03-06
update onboarding docs


jhanley634:
I still need to write up some ReadMe text which corresponds to the PR 90 + PR 102 code, and which shows the expected output so a newbie knows they're on the right track.

jhanley634:
Merged.

PR #108, opened 2025-03-09
103 update onboarding docs

jhanley634:
This PR mostly adjusts comment + *.md documentation to reflect recent code changes, including Docker support, closing out Issue #103. It also addresses a leftover Requested Change from PR 101, and offers browser access to a persistently running container. Here is a per-file summary:

- Switch to a persistent `bin/webserver-entrypoint.sh`, which runs until CTRL/C. (Put it in the background, or switch to a new Terminal.app tab to run new commands.)
- Listen to more than just localhost commands, so the container can export web results to laptop's browser. (Folks with security concerns can choose to spell out the "serve" command manually, without the "browse" shortcut.)
- Note that 3.9 testing reveals we need a modern interpreter. Comment on the "init" and "load" verbs introduced by a recent PR.
- Introduce new `bin/build-container.sh` instructions, so the human-readable instructions can be a one-liner, to simply invoke it.
- Introduce new `bin/webserver-entrypoint.sh`, which does init + load + browse
- Put docker-specific documentation in its own file. Also include example output, so folks know they're on the right track or so they can see which instruction died.

----

There's a design choice in here. We build up state that must be carried across {init, load, recommend}, and that state is slightly expensive, it costs several minutes. The entrypoint in this PR proposes to do that setup work and then keep the (otherwise transient!) container running indefinitely, preserving that state. So subsequent quick commands start `docker exec ...`.

There's at least three alternatives I'm willing to explore if people grumble, which would let quick commands start `docker run ...`.

1. Store state in the host's filesystem (the laptop's filesystem) using a docker volume -- this has the downside of "less isolation", "more details to juggle"
2. Store init + load state within the **image** (not the **container**) -- this has the downside that with `docker run cmd1` followed by `docker run cmd2`, cmd1 is ephemeral, so cmd2 won't see any of cmd1's FS side effects
3. Focus on mutating an external database, rather than a filesystem -- this has the downside of DBA tasks, TCP port configuration, DB configuration, complexity of perhaps having a single central database out in the cloud which multiple participants would need to coordinate making changes do


audiodude (approved):
LGTM

PR #111, opened 2025-03-13
Don't use header for non-human CSV

### LINKED ISSUE #110: Don't import CSV header row as data
Apparently the header of the CSV file is being imported as actual data:

```
sqlite> select * from rating where user_id = 'user_id';
user_id|movie_id|rating
```

The CSV shouldn't need a header, because it's not going to be read by humans.

audiodude:
Fixes #110


audiodude:
Superseded by #112

PR #112, opened 2025-03-14
110 write a failing test

jhanley634:
In working with various DB backends there's a pair of things I've become accustomed to:
- postgres COPY and mariadb LOAD FILE behave quite predictably (and fast)
- typically I'll want to pre-create a zero-row table with suitable schema, constraints, indexes

I mis-applied these lessons in a sqlite context. It turns out the documented behavior
of `.import` is quite different based on whether target table already exists or not.
I found this, ummm, surprising; it's a tool that may work today and bite you tomorrow.
This PR does a Root Cause Analysis and fixes my breakage, ensuring it never crops up again, by changing both test and target code.
It preserves the .CSV text header line, so Pandas and other programmatic consumers can treat the file as "self-describing".

----

Reviewer's guide:
- In the ETL target code, we do **not** pre-create the rating_csv table (so, alas, we can't specify that some fields are numeric, but that's fine).
- In the target code we `_validate_rating_table()` so this regrettable effect can't crop up in production.
- We display the CLI tool's stdout, which made no difference and proved convenient for debugging.
- We introduce "a failing (Red) test" which highlights how `.import` behaves differently according to whether destination table already exists. Plus there's a docs citation.
- Prefer to `run mb load` rather than test_etl(). Therefore the test method is deleted, and it cannot leave behind a "short" rating table.

### DIFF
--- a/mediabridge/data_processing/etl.py
+++ b/mediabridge/data_processing/etl.py
@@ -140,7 +140,7 @@ def _validate_rating_table() -> None:
     with Session(get_engine()) as sess:
         # We expect strictly numeric data here, not a column heading text label.
         q = sess.query(Rating).filter(Rating.user_id == "user_id")
-        assert 0 == len(list(q.all()))
+        assert 0 == len(list(q.all())), "Rating table contained header row"
 
 
 def _get_input_csv(max_rows: int, all_rows: int = 100_480_507) -> Path:
--- a/mediabridge/data_processing/etl.py
+++ b/mediabridge/data_processing/etl.py
@@ -115,7 +115,7 @@ def _insert_ratings(csv: Path, max_rows: int) -> None:
         conn.execute(text("DROP TABLE  IF EXISTS  rating_csv"))
         conn.execute(text(create_rating_csv))
         conn.commit()
-        _run_sqlite_child(
+        run_sqlite_child(
             [
                 ".mode csv",
                 ".headers on",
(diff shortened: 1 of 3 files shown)


mediabridge/data_processing/etl.py
```diff
@@ -139,6 +135,14 @@ def _insert_ratings(csv: Path, max_rows: int) -> None:
         # ETL finished in 468.062 s (eight minutes)
 
 
+def _validate_rating_table() -> None:
+    """Verifies that the code defect of Issue 110 never crops up again."""
+    with Session(get_engine()) as sess:
+        # We expect strictly numeric data here, not a column heading text label.
+        q = sess.query(Rating).filter(Rating.user_id == "user_id")
+        assert 0 == len(list(q.all()))
```
audiodude, line 143: Please put a descriptive error message after the assert.

`assert ..., 'Rating table contained header row'`

tests/recommender/two_movies_rec_test.py
```diff
@@ -24,24 +21,6 @@ class TestTwoMoviesRec(unittest.TestCase):
     # Ideally we should be able to have a user supply two movies they like,
     # and recommend a few more to them. We're not quite there yet.
 
-    def test_etl(self) -> None:
```
audiodude, line 27: Good riddance!

audiodude (approved):
LGTM

mediabridge/data_processing/etl.py
jhanley634: Ok. Sometimes I will think of the stack trace line number as "enough" for something I believe "cannot happen", and then will make it more diagnostic if it turns out that sometimes the impossible _does_ occasionally happen. But sure, happy to do it.
[110-write-a-failing-test a94da7b] Put diagnostic message after an assert

PR #113, opened 2025-03-20
Ollama API integration

JamesKohlsRepo:
Simple API calls to publically hosted ollama server


mediabridge/integrations/ollama_api.py
```diff
@@ -0,0 +1,40 @@
...
+    """Sends a prompt to the Ollama API and returns the response as plain text"""
+    payload = {"model": model, "prompt": prompt, "stream": False}
+    response = requests.post(url, headers=get_headers(), json=payload)
+    response.raise_for_status()  # raise an error if request fails
+    response_json = response.json()
+    return response_json.get("response", "")
+
+
+if __name__ == "__main__":
+    api_base_url = "https://ollama.tomato-pepper.uk/api"
```
audiodude, line 24: This is something of a security issue. From what I can tell, there is no password protecting this server. By checking in this code, you are potentially allowing untold millions (yes) of people on the internet to use your Ollama server and your computing resources. There are documented cases of people scraping GitHub to find unprotected API keys and server resources.

What we need is a way of making this URL secret, or password protecting it and making the passwords secret. I can go over that at the next meeting if you like, or we can discuss it otherwise.

tests/integrations/test_ollama_api.py
```diff
@@ -0,0 +1,27 @@
...
+        model_name = "llama3:latest"
+        input_prompt = "please say only the word 'Hello'"
+        response = generate_prompt_response(
+            model_name, input_prompt, f"{api_base_url}/generate"
+        )
+        self.assertEqual(response.strip(), "Hello")
+
+
+if __name__ == "__main__":
+    unittest.main()
```
jhanley634, line 27: Please rename this module to follow the convention used by the rest of the codebase:
ollama_api_test.py

I really like test_get_response(), as it tests what we care about. Nice, solid prompt. The underlying generate_prompt_response() could default a few things to DRY up this call site.

Consider renaming to a private `_get_headers`, which we don't explicitly test. The code here is simply repeating the implementation, which is never helpful, and super brittle.

nit: Consider deleting the very nice `__main__` guard + .main() call, as the module docstring eloquently points out that the usual calling convention is `python -m unittest`. (I tend to hit up-arrow RETURN on that a fair amount when doing an edit-debug cycle.) Including those two lines _can_ sometimes be helpful during initial test development, but seeing them replicated across N test modules grows a bit tedious.

mediabridge/integrations/ollama_api.py
```diff
@@ -0,0 +1,40 @@
...
+    for model in models_data:
+        print("-", model["name"])
+
+    model_name = "llama3:latest"
+    input_prompt = "please say only the word 'Hello'"
+
+    output_text = generate_prompt_response(
+        model_name, input_prompt, f"{api_base_url}/generate"
+    )
+    print("Model response:", output_text)
```
jhanley634, line 40: Nice annotations, thank you!

nit: `mypy --strict` would like to see a slightly longer signature of
`def get_headers() -> dict[str, str]:`

Consider improving the Public API design by defaulting some "boring", "seldom changed" parameters, so each call site can be more concise, specifying just the prompt:
```
def generate_prompt_response(prompt: str, model: str = "llama3", url: str = f"{api_base_url}/generate") -> str:
```
This is also suggesting that the module's Public API should always export `api_base_url`, not just within the `__main__` guard.

I am skeptical that `.get()` defaulting to empty string is a Good Thing. I suspect that `return response_json["response"]` would be better. If the "response" field is missing, it seems like we botched the [contract](https://en.wikipedia.org/wiki/Design_by_contract) and we _should_ `raise` an annoying KeyError, rather than paper it over and return an incorrect `str` result.

nit: Further to Travis's remarks about GETs from the wider internet, you minimally want to have a [logger](https://docs.python.org/3/library/logging.html) note the source IP and timestamp of incoming requests, so you can tell if there's trouble. And maybe the length of the response or (nearly the same thing) elapsed time to generate the response. I feel it's OK to merge this down, and then address the topic of "client secret" or "API token" in a subsequent PR.

nit: The `__main__` guard is very nice, but perhaps a little long. Consider banishing the code to `def main()`, leaving us with just a single guarded line, which invokes `main()`.

nit: Maybe with `typer.run(main)` we could accept random prompts as a [command line](https://pypi.org/project/typer) arg? (Yeah, I know, "creeping featurism!".)

nit: Pretty sure we don't need the ":latest" suffix on model names, as it defaults.

jhanley634 (approved):
The big thing I'm looking for is a file name rename prior to merging down, please. The rest is merely suggested "nice to haves", nothing that's really essential. There's more work that could be done on this, but I feel it would be outside the scope of PR 113. (I _would_ say outside the scope of work promised by Issue NNN, except there isn't one. Perhaps we'll open a new Issue for the next feature branch iteration.)

jhanley634:
Usually I try to find something to praise in an edit -- it tends to make subsequent criticisms come out as more constructive. However here, I feel that
> c486a63: Added Nginx Reverse proxy
is not an improvement, and should not be merged as-is to the `main` branch.

(I am not expert on the myriad ways that one might authenticate a client. But I have supported deployed systems through different growth phases of the lifecycle, including chatting with the FBI about a breach. If the goal is for this project to be a learning experience, I feel the project should be teaching Best practices rather than merely OK practices.)

We've already had some discussion about "uptime", "SLA", the effect of monitoring on MTTD + MTTR, and similar operational concerns. Let's talk about how a population of supported authenticated users interacts with DevOps.

Every deployed computer system is a complex system, because humans are part of the system, and they have a lot of details. The following two failure modes will happen with different frequencies, but in a long lived system both are inevitable:
1. If the server has a Secret, it _will_ be revealed, perhaps through carelessness, perhaps due to hacking, such as a zero-day exploit. Such exposure affects all N registered users.
2. Each year some fraction of the N registered users _will_ have a security event, such as losing whatever they stored their Secret password on, or being hacked or otherwise disclosing their credentials.

We can choose to adopt a design where the server has a single credential shared by all N users. Few designers would make that choice, as it is extremely disruptive to all N users when (not "if" -- "when!") a security incident happens. Typically item (1.) is simply excluded from the design space, and not even considered again.

Most designers will instead embrace item (2.), and shoulder the support cost of making N individual enrollments of registered users. Often there will be technical infrastructure, such as enrollment forms, to automate much of the flow. There's at least two techniques to implement that:
(A.) **shared secret**: Each user rolls e.g. a random 128-bit GUID, "securely" shares it with the server's list of authorized clients, and presents it over https: on each request. However, a server compromise would be trouble for all N users.
(B.) **public key**: Each user rolls e.g. an SSH {public, private} key pair, and discloses just the public part to the server. This very much simplifies enrollment and recovering from security incidents. Server never sees the secret part of a user's credentials.

Typically item (A.) is not a viable part of the design space, and a designer will prefer item (B.). I am not familiar with ways for an nginx config file to tackle (B.). I could definitely advise on how write Flask python code which authenticates PK credentials.

----

It was a (then "small") mistake to create the `llm-library` feature branch without having first created an Issue describing what we hoped to accomplish in that branch. You can't refer to the Issue to see what was intended, and I can't refer to it. I can't say that an authentication goal was or was not a part of the intended Use Case. I encourage you to first create an Issue prior to creating new feature branches. If you change your mind, hey, that's great, go back and add to the Issue, based on new stuff you've learned about the problem at hand.

----

tl;dr: Long ago I **approved** PR 113 and urged you to **merge as-is**, without listening to the siren song of feature creep. Yes, one could always add additional functionality to a proposed set of edits. But often the delay and uncertainty impacts the ability of other team members to build upon work that's already been accomplished. Let not the Perfect be the enemy of the Good.

I still advocate that the "old" edits, the "approved" edits, be merged down to `main`, even though Travis raises valid concerns about those edits.

I do **not** feel the current c486a63 edit should be merged. I believe it gives the appearance of "oh, we have solved the security problem", without doing a good job of surveying the literature, writing down the choices and decisions we have made, and then implementing them. Figuring out security details is a non-trivial problem. I feel it is out-of-scope for the "add features!" of PR 113, and would be better addressed in a subsequent PR.

audiodude:
@jhanley634 I've read your lengthy response and I wonder if you viewed the code change you were commenting on.

This does not add any kind of authentication system, basic or otherwise, to MediaBridge. It simply adds a way for MediaBridge to authenticate to an external system, which is an Ollama server running in @JamesKohlsRepo 's basement.

Now there are definitely operational issues concerning whether we want the MediaBridge "production" server, to rely on a server in someone's basement.

However from a security standpoint, users of *our* system do not need to authenticate to that server, so they don't need individual authentication methods, public key or otherwise.

I maintain that it would have been very unwise to expose James's Ollama server to the internet without any password or any security, and it was **not** appropriate to merge "as is".

jhanley634:
@audiodude, I read it, and I tried to execute it. There's a bash script (basically "curl") I have shared with James, which mirrors what the python code does. I verified it would `401` with the revised nginx config (which is not in source control, but that's a separate DevOps teachable moment), and verified that attempting to literally use the password "INPUT_PASSWORD_HERE" would of course fail. I infer there is an "enroll authorized user" step which does not yet appear in the ReadMe, and that at present I am not an authorized user. Which is fine. But it also impacts my ability to offer constructive criticism about proposed edits.

Security is a can of worms. We do not currently have an Issue which draws system boundaries around that can. Reasonable people will differ. Neglecting to write down the details makes it more likely that people will approach a PR with differing perspectives. We are hoping to reveal "things that could present rough edges to the process", and then help folks drive consensus toward a better, deployed, product.

audiodude:
@jhanley634 I thought more about what I wrote in light of what you wrote, and based on your last comment, I can see that we have a fundamentally different view of the "devops" of the system @JamesKohlsRepo is building.

I am viewing it as an "external" API that the MediaBridge service consumes. It sounds like you're thinking of it more as an internal part of the MediaBridge system. End users will not need to have an account or a password on the Ollama server. Only the MediaBridge server itself needs that password.

That said, I reconsidered your objections to having a single password (which is what this would have) and they make some sense in the regard that each developer will have a `ollama/secret_password` combination in their individual .env files. If any of the developer machines/files got compromised, it would necessitate changing that password for all users.

It is possible with HTTP Basic Auth to have multiple sets of credentials, so presumably we could provide credentials to each developer. Though fundamentally it creates a barrier for onboarding new developers.

Additionally, I have discussed with @JamesKohlsRepo offline the fact that HTTP Basic Auth is considered a fundamentally broken/insecure system.

Unfortunately, this might just be the facts of taking advantage of a physical computer with a GPU that is able to process low QPS, for the purposes of prototyping. Presumably we could swap in any number of LLMs-as-a-service when we wish to scale.

I still disagree that having Basic Auth is worse than not having it.

tests/integrations/test_ollama_api.py
JamesKohlsRepo: I believe these issues are addressed now

mediabridge/integrations/ollama_api.py
JamesKohlsRepo: I opted to remove main from ollama_api, it's cleaner as you mentioned. I tried my best to address most of the last nits

mediabridge/integrations/ollama_api.py
JamesKohlsRepo: Thankfully Ollama documented how it works with nginx. I rememberd sky talking about his reverse proxy, Im glad I could apply those principles to add basic auth. It's not really security, but it should keep out most basic bots or whatever

jhanley634:
> I still disagree that having Basic Auth is worse than not having it.

That's cool. We don't have agree to make forward progress. (I am large, I contain multitudes.)

I was proposing that PR 113 should be "small", introducing "new functionality", which uncovered a security detail that the feature branch had not previously considered. And I was proposing that a subsequent PR should tackle the security topic in a more careful manner than the c486a63 edit. That is, with N credentials for N developers, and ReadMe documentation on how to manage credentials. My very specific objection to adopting a single server-side shared secret across N parties is that it puts a strong social emphasis against reporting "oh, my laptop was compromised!" or "I accidentally posted the secret in a public channel". I have seen this play out in more than one social situation, involving people of differing aptitudes and levels of experience. When reporting a single "whoops!" security incident involves public shame and inconvenience for N-1 participants, that inevitably leads to under reporting of incidents. I'll have trouble embracing such an approach as a security "solution" until the DevOps details and assumptions are written down where the team can understand them and choose to adopt them.

PR #118, opened 2025-04-09
Update README.md

audiodude:
I use `yarn` in other projects, forgot that this project is just using `npm`.

Added instructions to install frontend deps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ Currently, we are only accepting contributions from members of the project who m
 
 This code uses Python 3. It is tested on Python 3.12, but will probably work on versions back to 3.10.
 
-To install the project dependencies, first install pipenv globally with `pip install pipenv`. Then create a virtual env/install dependencies with `pipenv install --dev`.
+To install the project Python dependencies, first install pipenv globally with `pip install pipenv`. Then create a virtual env/install dependencies with `pipenv install --dev`.
+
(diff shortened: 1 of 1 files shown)


JamesKohlsRepo (approved):
Sorry about approving the last commit. I'm make sure to run it myself in the future. I went through your instructions and it worked great, thanks for catching that mistake.

audiodude:
Oh no worries, no need to be hard on yourself, it was just a documentation change. And if you had yarn on your system like I do, it would have worked anyways.

jhanley634:
I went down a merry rabbit hole, attempting commands like these:
```
brew install yarn
rm -rf node_modules
yarn install
yarn add @vitejs/plugin-react --dev 
yarn add @babel/runtime --dev 
```
Spoiler alert: nope, didn't work. Now I see it's not expected to work.

-----

@cocomittens , I never ran the UI, due to lack of instructions in the ReadMe. I will take another stab at it later this week.

@audiodude , I had no idea there was an internet mongo server devoted to supporting this project, so I forked off a daemon on my laptop when trying to exercise callers of connect_to_mongo(), and strictly used an env var value of "localhost". (Besides, re-init and start-from-scratch seemed cleaner if it affects just a single project participant.) If there's a supported hostname to use, it should appear in the ReadMe. Currently `grep -i mongo README.md` reports zero rows. If there are access credentials involved, well that opens an additional can of worms...

PR #119, opened 2025-04-10
Load movie data into mongo based on sqlite database

audiodude:
Adds a command `mb load-mongo` that will load the netflix id, movie title, and year into a MongoDB "movies" collection.

There is currently a mongo instance running in the cloud that we can use, but it might be a problem if multiple developers are using it at once. As https://github.com/noisebridge/MediaBridge/pull/118#issuecomment-2794230879 points out, we need a better solution, potentially adding a docker-compose.yml that runs Mongo in a local container so that developers could connect to that instead.

For now, add `MONGODB_URI` to a `.env` file in the root directory.


JamesKohlsRepo (approved):
I like the modifications to etl. The load from sql and load_mongo functions look great. Looking forward to trying one I figure out mongo

PR #128, opened 2025-04-25
Add module for Flask backend API

audiodude:
This PR adds a new module `api`, that contains a basic Flask app. The app can be run with:

```
pipenv run mb serve
```

Appropriate instructions have also been added to the README.

Currently, the only implemented method is a movie title search that dumbly uses the SQL `LIKE` operator. If you've loaded your movie title data, and you're running the API server with the above command, you can try http://localhost:5000/api/v1/movie/search?q=Dinosaur


JamesKohlsRepo (approved):
well documented and well written, nice work!

PR #133, opened 2025-05-02
Feature/autocomplete UI

JamesKohlsRepo:
Migrated search bar logic to it's own file that fetches movies from the flask API. The search bar renders up to 10 movie titles, each of with hovering changes the color and reveals the release year. Clicking on a title populates the search bar

It could be improved with down arrow selection and submission, but I like where the search bar is now.

Screenshot 2025-05-01 at 21 39 28
Screenshot 2025-05-01 at 21 51 07


audiodude (approved):
This looks great! I told you writing an Autocomplete isn't as hard as it sounds! Definitely adding arrow navigation would be an improvement, but only if you're so motivated,

PR #134, opened 2025-05-03
Make search endpoint case insensitive

audiodude:
Just a quick fix so that the search endpoint is case insensitive. So now when you search `?q=dino` you'll get "Dinosaur Planet".


audiodude:
Feel free to merge if and when you approve!

JamesKohlsRepo (approved):
nice! looks good

Issue #135, opened 2025-05-05
add arrow key navigation to search


JamesKohlsRepo:
allow users to use the up and down arrow keys to quickly select movies, then press enter to add them to the movies list

PR #136, opened 2025-05-05
issue #135 adds arrow key navigation

JamesKohlsRepo:
Adds Arrow key navigation to suggestions, allow users to select content and press enter to add to list. Index starts at 0 so pressing enter will submit the first autocomplete item by default. 

May-04-2025 21-48-44

### DIFF
--- a/mediabridge-frontend/src/components/ui/searchbar.tsx
+++ b/mediabridge-frontend/src/components/ui/searchbar.tsx
@@ -16,6 +16,7 @@ const SearchBar = ({ title, setTitle, handleAddMovie, setSuggestions, suggestion
     const [isFocused, setIsFocused] = useState(false);
     const [highlightedIndex, setHighlightedIndex] = useState<number>(0);
 
+
     const fetchSuggestions = async (query: string) => {
         if (!query.trim()) return setSuggestions([]);
         try {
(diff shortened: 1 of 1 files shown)


mediabridge-frontend/src/components/ui/searchbar.tsx
```diff
@@ -46,24 +49,45 @@ const SearchBar = ({ title, setTitle, handleAddMovie, setSuggestions, suggestion
...
+                    {suggestions.map((movie, index) => (
                         <div
                             key={movie.id}
                             onClick={() => {
                                 setTitle(movie.title);
                                 setSuggestions([]);
                             }}
                             title={`Released in ${movie.year}`}
-                            className="px-4 py-2 hover:bg-gray-100 cursor-pointer text-left"
+                            className={`px-4 py-2 cursor-pointer text-left ${
```
audiodude, line 88: Nit: since this className is very complex and spans multiple lines, consider pulling it into it's own JS variable and referencing that:

```
const suggestionClassName = `px-4 py-2 cursor-pointer text-left ${
   index === highlightedIndex ? "bg-gray-200" : "hover:bg-gray-100"
}`

<div
   key={movie.id}
   ...
   className={suggestionClassNames}
  ...
```

mediabridge-frontend/src/components/ui/searchbar.tsx
```diff
@@ -46,24 +49,45 @@ const SearchBar = ({ title, setTitle, handleAddMovie, setSuggestions, suggestion
                     onKeyDown={(e) => {
                         if (e.key === "Enter") {
```
audiodude, line 50: nit: Basically the same comment as the other one. This "inline" function has become unwieldy and makes it hard to read where the components are. Consider extracting an onInputKeyDown function and using that here:

```
const onInputKeyDown = {(e) => {
                        if (e.key === "Enter") {
                            e.preventDefault();
                            if (highlightedIndex >= 0 && highlightedIndex < suggestions.length) {
                                const selected = suggestions[highlightedIndex];
                                handleAddMovie(selected);
                ...
};

...

<Input
  ...
  onKeyDown={onInputKeyDown}
  ...
/>
```

audiodude (approved):
This is great! Good work.

I left a few comments about possible refactors to help readability. Totally optional. I'll let you handle merging after you decide if you want to do them or not.

mediabridge-frontend/src/components/ui/searchbar.tsx
JamesKohlsRepo: sounds good. I do need to keep the variable in the map in order to get the index reference, but this will improve legibility for sure

mediabridge-frontend/src/components/ui/searchbar.tsx
JamesKohlsRepo: good call! I think the onKeyDown used to be a one liner but much functionality has been crammed in there haha

PR #138, opened 2025-05-05
137 restore annotations

jhanley634:
It should be the case that
- `pipenv run lint`, and
- `pipenv run test`
succeed on each commit that's proposed for a pull request. This PR restores that property. 

<s>Additionally it makes functions like `search_movies()` follow the usual Flask convention of appearing at module top level, so they are publicly available for other callers like unit tests to invoke them.</s>


audiodude (changes requested):
I appreciate the work to restore typing goodness and clean up some of the quirks of previous commits, so thank you!

However, I have to insist that you restore `mediabridge/api/main.py`, or at least restore it to a `create_app()` factory function pattern of Flask usage.

You can read more about it here: https://flask.palletsprojects.com/en/stable/patterns/appfactories/

Basically, it will allow us more flexibility and testing, and will become particularly important if and when we start using Flask blueprints.

jhanley634:
Hi, Travis.

I respectfully disagree. And by "respect" I mean that I restored that source file bit-for-bit, and then slightly munged it in order to lint clean. But honestly, I do not yet see the motivation. And I am willing to defer such discussion until we can meet in person at NoiseBridge.

At a high level, I tend to have a knee jerk response of "nested functions --> bad". Often the person using that technique hasn't thought through the "global variable" (ok, the "coupling") consequences of the nested namespace. But the really simple thing that falls out of "I defined nested foo()" is that, fine, 1st caller of foo() works great, but then defining a 2nd caller in some unit test module isn't going to go well, since it's hidden. An untestable function is likely one in which bugs are lurking. Guaranteed true? No. But I've seen it often enough, so I try to steer folks away from that.

On DARPA projects I always saw `app` defined at module top level, where others could `import` it, and routes defined for top level functions. At a medium-sized startup I defined or worked with a bunch of k8s flask microservices that put `app` and the routed functions at top level, which also worked well. One of those services relied on blueprints, written by a contractor who had moved on. It turned out the level of abstraction was not well supported, and in practice the dozen maintenance engineers who interacted with the code had trouble grasping how it should work and how the design should evolve. Lack of unit tests was part of the difficulty.

The cited URL mentions three Factory Improvements, which I'm willing to believe could be a good thing: {make..., call..., add...}. Our current codebase does not leverage such improvements, and generally does not convince me regarding "top level functions vs. nested". I invite subsequent PRs that continue to make the case!

audiodude (approved):
LGTM

PR #141, opened 2025-05-07
Run mypy in CI

audiodude:
Although we would like all of our code to have appropriate typings, as reported by `mypy`, and although there is currently a pipenv command to run that check (`pipenv run lint`), it is clear that developers will often not be paying attention to typings or forget to add them. And of course forget to run pipenv run lint. And probably not have `pre-commit` installed. Etc, etc, etc.

Here, we add a check to CI for mypy. If the lint script runs successfully, the check passes. If not, it fails, and the contributing developer must fix their PR before it can be merged.

This PR also includes changes to `lint.sh` to allow it to run `--only=mypy` (or "ruff" or "pyright"), which is mostly just used for CI.


.github/workflows/ci.yml
```diff
@@ -29,18 +29,38 @@ jobs:
...
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pipenv
+          pipenv install --dev
+
+      # Use mypy command directly from lint.sh script.
+      - name: Run MyPy
+        run: pipenv run lint -- --only=mypy
```
jhanley634, line 66: tl;dr: Approved.

minor grumbles / commentary:

The 3.11 interpreter seems slightly odd. The ReadMe explains

> This code uses Python 3. It is tested on Python 3.12, but will probably work on versions back to 3.10.

So I _guess_ I would have expected either (pessimal) 3.10, or (usual) 3.12? Or maybe a loop over versions .10 through .13 ? The ReadMe used to mention 3.9, until I verified that the codebase uses syntax which definitely won't work with that interpreter.

Also, I don't exactly get the `--only` motivation. I tend to think of a codebase in terms of "well, it lints clean, or it doesn't", similar to "the unit test suite is either Red, or it's Green". Maybe it's for more informative diagnostics, so we can better see _which_ aspect failed to lint? Maybe we desire separate pyright and mypy jobs which can pass or fail independently? Mostly I'm just saying that I wouldn't mind seeing additional comments or something in `doc/` describing the design considerations. It would aid a future maintenance engineer.

In lint.sh I kind of like `set -x` as it makes thing clearer when pasting symptoms into Discord or an LLM, but fine whatever. Of course what I really want is `make lint`, which is similarly clear, but that seems infeasible at the moment.

LGTM, ship it. :shipit:

jhanley634 (approved):
🚢

.github/workflows/ci.yml
audiodude: > tl;dr: Approved.
> 
> minor grumbles / commentary:
> 
> The 3.11 interpreter seems slightly odd. The ReadMe explains
> 
> > This code uses Python 3. It is tested on Python 3.12, but will probably work on versions back to 3.10.
> 
> So I _guess_ I would have expected either (pessimal) 3.10, or (usual) 3.12? Or maybe a loop over versions .10 through .13 ? The ReadMe used to mention 3.9, until I verified that the codebase uses syntax which definitely won't work with that interpreter.

It's just a copy paste job. I can change it to 3.12, np.

> Also, I don't exactly get the `--only` motivation. I tend to think of a codebase in terms of "well, it lints clean, or it doesn't", similar to "the unit test suite is either Red, or it's Green". Maybe it's for more informative diagnostics, so we can better see _which_ aspect failed to lint? Maybe we desire separate pyright and mypy jobs which can pass or fail independently? Mostly I'm just saying that I wouldn't mind seeing additional comments or something in `doc/` describing the design considerations. It would aid a future maintenance engineer.

The motivation is that I don't want to run pyright right now, and when I do, I want ruff/mypy/pyright to be separate "checks" in CI, so that someone can see "Oh, mypy failed" instead of "Oh, the comprehensive lint of many steps failed".

The other motivation is that ruff can provide inline error messages in github with it's `--output_format=github` flag. But we don't want that in normal linting. So this again points to having each linter be it's own check, and additionally to not run the ruff linter from a shell script.

> In lint.sh I kind of like `set -x` as it makes things clearer when pasting symptoms into Discord or an LLM, but fine whatever. Of course what I really want is `make lint`, which is similarly clear, but that seems infeasible at the moment.

+x is the quick and dirty way to get bash to print commands it is running. Unfortunately, with all the extra `if...fi` stuff, it will print all of that, plus the getopts call. So then it's just a bunch of junky logspam when you run the script. I think this way is good because I'm echoing the precise command that's being run, so it's maintainable at least.

> 
> LGTM, ship it. :shipit:

:+1:

PR #143, opened 2025-05-07
Use the Flask-SQLAlchemy plugin for flask

audiodude:
The main advantage of using this plugin/extension is that it automatically handles opening and closing database connections as part of the request lifecycle. So the database connection is available using the `db` object whenever we are inside of a Flask "application context" (https://flask.palletsprojects.com/en/stable/appcontext/).

This PR also takes advantage of the App Factory pattern introduced in the original API PR to allow for configuration of the app. By default, `config_name` is `None`, so we use the development config (or whatever is specified in the `FLASK_ENV` environment variable). For testing, we can explicitly call `create_app('testing')` and get an in-memory sqlite database for testing. This is advantageous because we don't have to worry about "creating"/"cleaning up" a sqlite database before/after tests.

Note that this would not be possible if the Flask `app` object were defined at the top level of the module, because `db.init_app()` would be called when the module was imported and we'd have no opportunity to configure it.


JamesKohlsRepo:
These changes are great! Should make mocking the database for front-end testing easier. I believe we discussed using Cypress before. Though the more standard and faster method seems to be Vitest/Jest and msw

JamesKohlsRepo (approved):
awesome work!

audiodude:
> These changes are great! Should make mocking the database for front-end testing easier. I believe we discussed using Cypress before. Though the more standard and faster method seems to be Vitest/Jest and msw. What do you think is the best approach?

It's unrelated to Flask, which testing framework we use for the frontend. The Flask app exists only in Python, so will always be tested in pure Python.

As for frontend testing, if you're going to implement it, feel free to use whatever you like! I only suggested Cypress because that's what I have the most experience with.

PR #144, opened 2025-05-07
Add mypy config to pyproject.toml

audiodude:
In a future PR, presumably after #141 has landed, we can clean up `lint.sh` to remove the redundant options.

### DIFF
--- a/bin/lint.sh
+++ b/bin/lint.sh
@@ -48,7 +48,7 @@ fi
 
 # If there is no value for "--only", or there is and it equals "mypy"
 if [[ -x "$only_value" || "$only_value" == "mypy" ]]; then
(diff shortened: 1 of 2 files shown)


jhanley634 (approved):
Looks good.

But please delete the "show_traceback = true" line, as I don't feel I can justify it. I mean, I think it's harmless. I think I tacked it on in response to this diagnostic: "note: please use --show-traceback to print a traceback when reporting a bug". I have no reason to believe it improves the linting of the current codebase. One can always copy-n-paste the mypy command and tack on --show-traceback if it seems relevant to a future issue. When it comes to configs I feel that "simpler is better than complex", and accepting defaults as much as possible reduces the burden on future maintenance engineers.

audiodude:
Did the lint cleanup here, since I've merged #141. It would be a merge nightmare if I tried to do it before then.

PR #145, opened 2025-05-07
Add mypy to recommended VSCode extensions

audiodude:
Quick PR to recommend a mypy extension for VSCode. I found this very helpful for debugging and fixing mypy errors, now that clean mypy is going to be required in #141.


JamesKohlsRepo (approved):
perfect we love to see it!

PR #146, opened 2025-05-07
Flask tests

audiodude:
Write the first tests for the Flask app, using pytest fixtures. Hopefully this demonstrates how easy testing can be, once it is set up properly. It can be difficult to understand all of the options for pytest with fixtures, but this basic setup is all that we really need. The key is that each fixture (`app` and `client`) are `yield`ed to the required functions, which means that the test functions operate in the same context (like "Context Manager" context) as was active when the yield was called.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,5 +5,6 @@
     "editor.codeActionsOnSave": {
       "source.organizeImports": "explicit"
     }
-  }
+  },
+  "editor.indentSize": "tabSize"
 }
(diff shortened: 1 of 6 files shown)


JamesKohlsRepo (approved):
sweet! With those flask changes mocking and testing api queries looks great. What is a test client in the context of conftest?

audiodude:
> sweet! With those flask changes mocking and testing api queries looks great. What is a test client in the context of conftest?

A test client is just a utility that Flask provides for testing. It's a way of making simulated HTTP requests to an endpoint, specified by its full URL, without having to actually use HTTP. It also provides some helpers for doing things like making POST requests without having to craft urlencoded parameters.

PR #149, opened 2025-05-13
Basic Cypress Implementation

JamesKohlsRepo:
One test to see if website loads, if this looks fine then next will be mocking data and adding/removing movies I think

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,12 @@ To run unittests:
 pipenv run test
 ```
 
+To run Cypress:
+
+```bash
+npx cypress open
+```
+
 These tests are also evaluated via a GitHub action when opening or updating a PR and must pass before merging.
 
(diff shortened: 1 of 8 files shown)


audiodude (changes requested):
This is fine, but please add cypress to package.json and include instructions in README.md for how to run the frontend tests. I'll take care of adding a github action for them.

audiodude (approved):

JamesKohlsRepo:
I do have some questions about the next steps. I'm mocking the data by creating a movies.json in cypress/fixtures. then mocking the api call by returning movies.json. The issue is because we're returning static data, the autocomplete does not work and we always see everything. Is there a way to mock the database more accurately?

audiodude:
I'll respond on the issue.

PR #155, opened 2025-05-22
154 fix lint sh

jhanley634:
Script never ran any linters before, but now it does.


Pipfile
```diff
@@ -8,7 +8,7 @@ beartype = "~=0.19"
 datasette = "~=0.65"
 flask = "~=3.1.0"
 flask-cors = "~=5.0"
-flask-sqlalchemy = "~=3.1.1"
+flask_sqlalchemy = "~=3.1.1"
```
audiodude, line 11: why does this cause a problem but not `flask-cors` above?

audiodude (approved):
LGTM

PR #158, opened 2025-07-01
Add recommend endpoint

cocomittens:
# Add recommend button
Adds recommendation endpoint to the backend & calls it when clicking the "Get Recommendations button" on the frontend. Functionality only, shows a list of the titles of the retrieved recommendations (no styling yet). Will refine further, just putting this up for review to ensure it's on the right track & using the correct recommendation function.

## Screenshots
image

### DIFF
--- a/mediabridge-frontend/src/pages/SelectMovies.tsx
+++ b/mediabridge-frontend/src/pages/SelectMovies.tsx
@@ -1,6 +1,7 @@
 import MovieList from "../components/MovieList";
 import MovieSearch from "../components/MovieSearch";
 import { Movie } from "../types/Movie.ts";
+import { useState } from "react";
 
 const SelectMovies = ({
   movies,
@@ -11,19 +12,30 @@ const SelectMovies = ({
   addMovie: (movie: Movie) => void;
   removeMovie: (id: string) => void;
 }) => {
+  const [recommendations, setRecommendations] = useState<string[]>([]);
+  console.log(recommendations);
+
   return (
     <div className="min-h-screen flex flex-col">
       {/* Top full-width section */}
       <div className="w-screen">
         <div className="max-w-screen-mx flex justify-center">
-          <MovieSearch movies={movies} addMovie={addMovie} />
+          <MovieSearch movies={movies} addMovie={addMovie} setRecommendations={setRecommendations} />
         </div>
       </div>
 
       {/* Content section below */}
-      <div className="px-4 py-8">
+      {
+        recommendations.length > 0 ?
+        <div>
+          {recommendations.map(recommendation => (<div key={recommendation}>{recommendation}</div>))}
+          </div>
+        :
+        <div className="px-4 py-8">
         <MovieList movies={movies} removeMovie={removeMovie} />
       </div>
+      }
+
     </div>
   );
 };
--- a/mediabridge/api/app.py
+++ b/mediabridge/api/app.py
@@ -51,22 +51,14 @@ def create_app(config_name: str | None = None) -> Flask:
             movies_list = [row._asdict() for row in movies]
             return jsonify(movies_list), 200
 
-    @app.route("/api/v1/movie/recommend")  # type: ignore
+    @app.route("/api/v1/movie/recommend", methods=["POST"])  # type: ignore
     def recommend_movies() -> tuple[Response, int]:
-        query = request.args.get("q")
-        if not query:
-            return jsonify({"error": "Query parameter 'q' is required."}), 400  # type: ignore
-
-        with db.engine.connect() as conn:
-            pattern = f"%{query}%"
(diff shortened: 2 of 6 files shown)


audiodude:
As far as the styling goes, can't you just reuse `MovieItem` and hide the remove button?

JamesKohlsRepo (commented):
Great to see your commit! I tired to keep the style similar to how you initially designed it. It would be great to have your help to clean up the styling and less then optimal design choices I've made.
For the Ruff error, I think you can fix line 26 by removing the double quotations: 
return f"{movie.get('netflix_id')}"

mediabridge/api/app.py
```diff
@@ -51,6 +52,23 @@ def search_movies() -> tuple[Response, int]:
             movies_list = [row._asdict() for row in movies]
             return jsonify(movies_list), 200
 
+    @app.route("/api/v1/movie/recommend", methods=["POST"])  # type: ignore
```
audiodude, line 55: This can be a GET, not a POST, because it doesn't modify any data.

mediabridge/api/app.py
```diff
@@ -51,6 +52,23 @@ def search_movies() -> tuple[Response, int]:
...
+    def recommend_movies() -> tuple[Response, int]:
+        data = request.get_json()
+        if not data or "movies" not in data:
+            return jsonify({"error": "JSON body with 'movies' array is required."}), 400
+        movies = data["movies"]
+        if not isinstance(movies, list) or not all(isinstance(m, str) for m in movies):
```
audiodude, line 61: We originally discussed sending in the movies as IDs, not titles. Check with @JamesKohlsRepo if we still want to do that.

mediabridge/api/app.py
```diff
@@ -51,6 +52,23 @@ def search_movies() -> tuple[Response, int]:
...
+            return jsonify({"error": "'movies' must be a list of strings."}), 400
+
+        try:
+            rec_ids = recommend()
```
audiodude, line 65: As discussed, no worries that this doesn't actually recommend anything for now.

mediabridge/api/app.py
```diff
@@ -51,6 +52,23 @@ def search_movies() -> tuple[Response, int]:
...
+            rec_titles = [get_title(int(mid)) for mid in rec_ids]
+        except Exception as e:
```
audiodude, line 67: It's considered bad practice to `except Exception`, because if there are logical errors in your code (`NameError`, `IndexError`, etc), they will also get caught here. Prefer to catch the precise exception you're worried about.

mediabridge-frontend/src/pages/SelectMovies.tsx
```diff
@@ -11,21 +12,31 @@ const SelectMovies = ({
...
         <div className="max-w-screen-mx flex justify-center">
-          <MovieSearch movies={movies} addMovie={addMovie} />
+          <MovieSearch movies={movies} addMovie={addMovie} setRecommendations={setRecommendations} />
         </div>
       </div>
 
       {/* Content section below */}
-      <div className="px-4 py-8">
+      {
+        recommendations.length > 0 ?
```
audiodude, line 28: Maybe you can just use MovieItem here? Or MovieList without the `removeMovie` property? Then the `MovieList` component could check if the optional `removeMovie` function was passed in, and if not, hide the remove buttons.

audiodude (changes requested):

cocomittens:
Updated the PR to use movie IDs instead of titles and use the MovieList component, let me know what yall think

mediabridge/api/app.py
```diff
@@ -52,22 +52,35 @@ def search_movies() -> tuple[Response, int]:
...
+        with db.engine.connect() as conn:
+            movie = conn.execute(
+                text("SELECT * FROM movie_title WHERE id = :id"),
+                {"id": movie_id},
+            ).fetchone()
+            if not movie:
+                return jsonify({"error": "Movie not found"}), 404
+            return jsonify(dict(movie._mapping)), 200
+
+    @app.route("/api/v1/movie/recommend", methods=["GET"])  # type: ignore
```
audiodude, line 66: nit: `methods=["GET"]` is not required, that's the default.

audiodude (commented):
Resolved some comments and added another one.

jhanley634:
I'm not certain, but I _think_ so-called "smart quotes" are causing a merge conflict. Note the four bold **e2 80 9x** sequences in the attached PDF.

The lockfile conflicts are pretty vanilla. Just "accept theirs", or regenerate with `pipenv lock --dev`, to resolve. It may be convenient to do this with five minutes of pair programming at NoiseBridge.

[hex_dump.pdf](https://github.com/user-attachments/files/21398814/hex_dump.pdf)

PR #159, opened 2025-07-01
Fix ruff and bump Python version

### LINKED ISSUE #160: Ruff is failing in CI
According to the Ruff docs, there is a `target-version` parameter (https://docs.astral.sh/ruff/settings/#target-version). In our case, it seems to be set by our `pyproject.toml`:

https://github.com/noisebridge/MediaBridge/blob/main/pyproject.toml#L5

We're saying that we support Python back to 3.10, which results in these errors:

https://github.com/noisebridge/MediaBridge/actions/runs/16010999805/job/45168584029?pr=159

However, the thing that I don't understand (and the real problem IMO) is that I don't see the errors when I run ruff on the command line:

```
tmoney-macbook:MediaBridge tmoney$ pipenv run ruff check .
All checks passed!
```

I attempted to fix it with #159 by making the CI run ruff the same way I do on my command line, but it didn't work.

audiodude:
Fixes #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ Currently, we are only accepting contributions from members of the project who m
 
 ## Development
 
-This code uses Python 3. It is tested on Python 3.12, but will probably work on versions back to 3.10.
+This code requires Python 3.12.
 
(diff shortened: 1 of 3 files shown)


JamesKohlsRepo (approved):
lgtm!

PR #168, opened 2025-07-22
167 pipfile maint

jhanley634:
Uprev package deps.
Also minor type checking, and lint edits, with no functional changes.


audiodude (approved):
This is fine, but I don't think it should be necessary to manual update dependencies like `~=2.2` to `~=2.3` right? If an available version >2.2 and < 3.0 is available, doesn't the former already allow Pipenv the ability to upgrade us automatically? It should just be a matter of regenerating the lock file.

